PR #128, opened 2015-01-06
Use getDistributedNodes to get nodes distributed into shadowRoots

### LINKED ISSUE #126: False Negatives running in chrome browser, but not running through phantomjs
When we do `axs.Audit.run()` to Google.com as in https://github.com/GoogleChrome/accessibility-developer-tools/issues/125 we get a few `FAIL`s, but if we run the same script in google.com in Chrome's console we do not:

screen shot 2014-12-23 at 4 56 14 pm

This may be relevant and high priority for the Chrome Extension.

alice:
Fixes #126

@RickSBrown PTAL
I thought I might be able to submit a PR to your PR, but it turns out you can only send PRs to the repo you forked from. Seems to miss the point of git a bit! Anyway, your original change is in there.

### DIFF
--- a/src/js/AuditRule.js
+++ b/src/js/AuditRule.js
@@ -134,55 +134,49 @@ axs.AuditRule.collectMatchingElements = function(node, matcher, collection,
     if (element && matcher.call(null, element))
         collection.push(element);
 
-    // Descend into node:
-    // If it has a ShadowRoot, ignore all child elements - these will be picked
-    // up by the <content> or <shadow> elements. Descend straight into the
-    // ShadowRoot.
+    // Descend into node
     if (element) {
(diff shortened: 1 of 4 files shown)


RickSBrown:
Great stuff. I'll admit I had to do a crash course on shadow DOM to understand how this works. This is definitely a better approach than I had originally proposed since it correctly ignores "light DOM" nodes that are not part of the "rendered DOM" (remember when there was just "The DOM"?).

alice:
Yeah, I had to do the crash course last year (or was it the year before?) when I wrote the original code! Luckily there's a bunch of good content about it on html5rocks.com and of course I can go bug my colleagues who work on it all day if I don't understand something.

Issue #133, opened 2015-01-09
Detect and warn if we reuse the same code for more than one rule.


RickSBrown:
For example if two rules try to register themselves with `code: 'AX_TITLE_01'` then the codebase should complain (at a glance probably belongs here: `axs.AuditRules.getRule`).

PR #135, opened 2015-01-10
Tighten up the mechanisms for registering/accessing audit rules.

RickSBrown:
Prevent audit rules being registered with duplicate name or code (Closes issue #133).
Added new method`axs.AuditRules.addRule` to register rules instead of adding them to `axs.AuditRule.specs`.
Removed `axs.AuditRule.specs` - this breaks backwards compatibility - the extension will need to be updated.
Removed public `axs.AuditRules.rules` because:
    a) it was never used;
    b) there is a getter method to access these.
Added new method `axs.AuditRules.getRules` to replace the functionality of `axs.AuditRule.specs` and axs.AuditRules.rules.

Fixed duplicate AX_ARIA_04
Fixed duplicate AX_TITLE_01

### DIFF
--- a/src/js/AuditRules.js
+++ b/src/js/AuditRules.js
@@ -16,21 +16,46 @@ goog.require('axs.AuditRule');
 
 goog.provide('axs.AuditRules');
 
-/**
- * Gets the audit rule with the given name.
- * @param {string} name
- * @return {axs.AuditRule}
- */
-axs.AuditRules.getRule = function(name) {
-    if (!axs.AuditRules.rules) {
-        /** @type Object.<string, axs.AuditRule> */
-        axs.AuditRules.rules = {};
-        for (var specName in axs.AuditRule.specs) {
-            var spec = axs.AuditRule.specs[specName];
-            var auditRule = new axs.AuditRule(spec);
-            axs.AuditRules.rules[spec.name] = auditRule;
-        }
-    }
+(function(){
+    var auditRulesByName = {};
+    var auditRulesByCode = {};
 
-    return axs.AuditRules.rules[name];
-};
+    /**
+     * Instantiates and registers an audit rule.
+     * If a conflicting rule is already registered then the new rule will not be added.
+     * @param {axs.AuditRule.Spec} spec The object which defines the AuditRule to add.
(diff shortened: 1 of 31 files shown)


RickSBrown:
@alice PTAL

FYI this breaks backwards compatibility with the extension but I'll make the changes there too.

src/js/AuditRules.js
```diff
@@ -16,21 +16,46 @@ goog.require('axs.AuditRule');
...
+     * If a conflicting rule is already registered then the new rule will not be added.
+     * @param {axs.AuditRule.Spec} spec The object which defines the AuditRule to add.
+     * @throws {Error} If the rule duplicates properties that must be unique.
+     */
+    axs.AuditRules.addRule = function(spec) {
+        // axs.AuditRule.specs[spec.name] = spec;  // This would add backwards compatibility
+        // create the auditRule before checking props as we can expect the constructor to perform the
+        // first layer of sanity checking.
+        var auditRule = new axs.AuditRule(spec);
+        if (auditRule.code in auditRulesByCode || auditRule.name in auditRulesByName)
```
alice: How about splitting these out into two separate cases so we can have two separate error messages?

src/js/AuditRules.js
```diff
@@ -16,21 +16,46 @@ goog.require('axs.AuditRule');
...
+     * @param {string} name The name (or code) of an audit rule.
+     * @return {axs.AuditRule}
+     */
+    axs.AuditRules.getRule = function(name) {
+        return auditRulesByName[name] || auditRulesByCode[name] || null;
+    };
+
+    /**
+     * Gets all registered audit rules.
+     * @param {boolean=} namesOnly If true then the result will contain only the rule names.
```
alice: This may be inconsistent in the codebase, but I try to always give optional parameters an opt_ prefix, i.e. opt_namesOnly, to be extra clear.

alice:
:+1: Go ahead and pull. Should I expect a PR on the extension as well?

RickSBrown:
Sweet.
Yes, I've started looking at the extension - I should get a chance in the next few days.

Issue #130, opened 2015-01-09
Reference "applied role" in axs.utils.getRoles


RickSBrown:
The ARIA spec allows for multiple roles but only one is applied.
`axs.utils.getRoles` should indicate which role was "applied" as this is likely to be required in many cases.

For example given the element below:

`<span role="foo bar checkbox button">`

`axs.utils.getRoles` should indicate that the applied role is `checkbox`.

This is best implemented as a new property `applied` on the return value which references the applied role `roleObject`.

RickSBrown:
This was closed by PR#138

Issue #131, opened 2015-01-09
Fix "valid" flag on return value of axs.utils.getRoles


RickSBrown:
Currently the return value of `axs.utils.getRoles` will indicate `valid = false` if any role is invalid.
This should be changed so that `valid` is only false if there are no valid roles and one or more invalid roles. In other words, if there is a valid role then the `valid` flag must be `true`.

Note: need to ensure the extension continues to flag invalid roles after this change has been made.

RickSBrown:
This was closed by PR#138

PR #138, opened 2015-01-15
Updates to axs.utils.getRoles and related changes to some Audits.

RickSBrown:
Closes issue #130 and #131.

### DIFF
--- a/src/audits/AriaRoleNotScoped.js
+++ b/src/audits/AriaRoleNotScoped.js
@@ -37,8 +37,8 @@ axs.AuditRules.addRule({
         var elementRole = axs.utils.getRoles(element);
         if (!elementRole || !elementRole.applied)
             return false;
-        elementRole = elementRole.applied;
-        var ariaRole = elementRole.details;
+        var appliedRole  = elementRole.applied;
+        var ariaRole = appliedRole.details;
         var requiredScope = ariaRole['scope'];
         if (!requiredScope || requiredScope.length === 0) {
             return false;
@@ -47,8 +47,8 @@ axs.AuditRules.addRule({
         while ((parent = parent.parentNode)) {
             var parentRole = axs.utils.getRoles(parent, true);
             if (parentRole && parentRole.applied) {
-                parentRole = parentRole.applied;
-                if (requiredScope.indexOf(parentRole.name) >= 0)  // if this ancestor role is one of the required roles
+                var appliedParentRole = parentRole.applied;
+                if (requiredScope.indexOf(appliedParentRole.name) >= 0)  // if this ancestor role is one of the required roles
                     return false;
             }
         }
--- a/src/js/AccessibilityUtils.js
+++ b/src/js/AccessibilityUtils.js
@@ -1055,8 +1055,8 @@ axs.utils.getRoles = function(element, implicit) {
     for (var i = 0; i < roleNames.length; i++) {
         var role = roleNames[i];
         var ariaRole = axs.constants.ARIA_ROLES[role];
-        var roleObject = {'name': role};
-        if (ariaRole && !ariaRole['abstract']) {
+        var roleObject = { 'name': role };
+        if (ariaRole && !ariaRole.abstract) {
             roleObject.details = ariaRole;
             if (!result.applied) {
                 result.applied = roleObject;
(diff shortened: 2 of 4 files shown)


RickSBrown:
@alice PTAL

src/audits/AriaRoleNotScoped.js
```diff
@@ -35,11 +35,9 @@ axs.AuditRules.addRule({
          * Checks that this element is in the required scope for its role.
          */
         var elementRole = axs.utils.getRoles(element);
-        if (!elementRole || !elementRole.roles.length)
-            return false;
-        elementRole = elementRole.roles[0];
-        if (!elementRole || !elementRole.valid)
+        if (!elementRole || !elementRole.applied)
             return false;
+        elementRole = elementRole.applied;
```
alice: I realise this aspect isn't new here, but I'm still not comfortable with reassigning `elementRole` here. Could we create a new var instead, e.g. `appliedRole`?

src/audits/AriaRoleNotScoped.js
```diff
@@ -48,8 +46,8 @@ axs.AuditRules.addRule({
         var parent = element;
         while ((parent = parent.parentNode)) {
             var parentRole = axs.utils.getRoles(parent, true);
-            if (parentRole && parentRole.roles.length) {
-                parentRole = parentRole.roles[0];
+            if (parentRole && parentRole.applied) {
+                parentRole = parentRole.applied;
```
alice: Similarly, something like `appliedParentRole` perhaps?

src/js/AccessibilityUtils.js
```diff
@@ -1051,22 +1051,25 @@ axs.utils.getRoles = function(element, implicit) {
...
-    var roles = [];
-    var valid = true;
+    var result = { roles: [], valid: false };
     for (var i = 0; i < roleNames.length; i++) {
         var role = roleNames[i];
         var ariaRole = axs.constants.ARIA_ROLES[role];
-        if (ariaRole && !ariaRole.abstract) {
-            var roleObject = {'name': role, 'details': axs.constants.ARIA_ROLES[role], 'valid': true};
-            roles.push(roleObject);
+        var roleObject = {'name': role};
```
alice: Space between { and content and } for object literal.

src/js/AccessibilityUtils.js
```diff
@@ -1051,22 +1051,25 @@ axs.utils.getRoles = function(element, implicit) {
...
+        if (ariaRole && !ariaRole['abstract']) {
```
alice: Why did this change to bracket notation?

test/js/utils-test.js
```diff
@@ -307,3 +311,65 @@ test("getRoles on element with abstract role.", function() {
         }
     }
 });
+
+test("getRoles on element with multiple valid roles.", function() {
+    var roles = ['checkbox', 'button', 'radio'];
+    var expectedRoles = roles.map(function(role){
```
alice: space in "function(role) {"

test/js/utils-test.js
```diff
@@ -307,3 +311,65 @@ test("getRoles on element with abstract role.", function() {
...
+    };
+    var element = document.createElement('div');
+    element.setAttribute('role', roles.join(' '));
+    var actual = axs.utils.getRoles(element);
+    deepEqual(actual, expected);
+});
+
+test("getRoles on element with invalid and valid roles.", function() {
+    var roles = ['foo', 'button', 'bar'];
+    var expectedRoles = roles.map(function(role){
```
alice: What do you think about scoping these two tests in an IIFE and pulling this function out into a named local helper function?

alice:
Sorry for being so ridiculously slow on this. Looks mostly good, just a couple of minor issues. Fine to pull once those are addressed.

src/js/AccessibilityUtils.js
RickSBrown: In days of yore `abstract` was a "future reserved word" and the bracket notation would "hide" it since it is just a string. I actually don't remember why I changed it, I must have been having a flashback. I'll put it back how it was since [it is no longer reserved](http://www.ecma-international.org/ecma-262/5.1/#sec-7.6.1.2).

test/js/utils-test.js
RickSBrown: I think that's a mighty fine idea. Shall do.

src/js/AccessibilityUtils.js
alice: Ahhh thanks for the explanation!

PR #139, opened 2015-01-26
Fixing audio without controls rule and adding a test for it.

### LINKED ISSUE #25: Also validate scope if it's a DOM node
In the rules, querySelectorAll is used, but that only selects descendants, not root element as well. I think we should validate the root element too, if it's a DOM node (not the document of course).

yurypetrov:
Fixes #25.

### DIFF
--- a/test/audits/audio-without-controls-test.js
+++ b/test/audits/audio-without-controls-test.js
@@ -1,0 +1,14 @@
+module("Audio");
+
+test("Audio has no controls attribute.", function() {
+  // Setup fixture
+  var fixture = document.getElementById('qunit-fixture');
+  fixture.innerHTML = "<audio src=\"http://developer.mozilla.org/@api/deki/files/2926/=AudioTest_(1).ogg\" autoplay></audio>";
+  var audio = document.querySelector('audio');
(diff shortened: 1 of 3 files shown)


yurypetrov:
@alice Please take a look.

test/audits/audio-without-controls-test.js
```diff
@@ -0,0 +1,14 @@
+module("Audio");
+
+test("Audio has no controls attribute.", function() {
+  // Setup fixture
+  var fixture = document.getElementById('qunit-fixture');
+  fixture.innerHTML = "<audio src=\"http://developer.mozilla.org/@api/deki/files/2926/=AudioTest_(1).ogg\" autoplay></audio>";
+  var audio = document.querySelector('audio');
+  audio.addEventListener('canplay', function() {
```
alice: As discussed, we need to come up with a way to block this test finishing until this assertion has run. Check out the QUnit doc: http://api.qunitjs.com/

ckundo:
@yurypetrov @alice did you find a way to properly test this?

Diogolaz12:
test/index.html

PR #145, opened 2015-04-21
updates links to Audit Rules wiki to use the new anchors

jaimeiniesta:
Hello, I recently updated the [Audit Rules](https://github.com/GoogleChrome/accessibility-developer-tools/wiki/Audit-Rules) wiki page to simplify the anchors on every rule, so they are easier to link from other projects ([see an example](https://github.com/accesslint/access_lint)).

I think that using just the audit rule ID, for example `#ax_aria_12` is much clearer than using a longer English string like `#-ax_aria_12--this-element-does-not-support-aria-roles-states-and-properties`.

The description text, "this element does not support aria roles states and properties" is much more prone to change in the future than the ID "ax_aria_12". Also, the anchors included a whitespace at the beginning that made them look a bit weird.

This PR changes the links to every audit rule to use the new anchors.

### DIFF
--- a/src/audits/AriaOnReservedElement.js
+++ b/src/audits/AriaOnReservedElement.js
@@ -22,7 +22,7 @@ goog.require('axs.properties');
 axs.AuditRules.addRule({
     name: 'ariaOnReservedElement',
     heading: 'This element does not support ARIA roles, states and properties',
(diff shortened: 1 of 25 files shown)


jaimeiniesta:
I signed it!

alice:
Fantastic, thanks @jaimeiniesta!

PR #146, opened 2015-04-21
Remove extraneous externs

alice:
@ckundo PTAL


alice:
@googlebot You're hilariously way off here.

ckundo:
@alice LGTM, pending a passing build!

Issue #144, opened 2015-04-15
Building project => error: Variable "x" first declared in externs.zip//html5.js


kristapsmelderis:
info: '--force' allows to proceed further

log:
Warning: ./src/js/externs/externs.js:32: WARNING - Redeclared variable: HTMLContentElement
function HTMLContentElement() {}
         ^

./src/js/externs/externs.js:43: WARNING - Redeclared variable: HTMLShadowElement
function HTMLShadowElement() {}
         ^

./src/js/externs/externs.js:32: ERROR - Variable HTMLContentElement first declared in externs.zip//html5.js
function HTMLContentElement() {}
         ^

./src/js/externs/externs.js:43: ERROR - Variable HTMLShadowElement first declared in externs.zip//html5.js
function HTMLShadowElement() {}

kristapsmelderis:
I did a "bad" thing and commented out the above mentioned code from externs.js and after that the project was building successfully, new axs_test.js was generated.

With the above mentioned error (not commenting the code out) during the build, the new audit file is not being generated.

jaimeiniesta:
I have similar errors trying to build it (master branch, OS X Yosemite):

```bash
$ grunt --version
grunt-cli v0.1.13
grunt v0.4.5

$ npm --version
2.7.5

$ grunt
Running "save-revision" task

Running "git-describe" task
Git Revision: 7339ac43b963264ef74b69458aefebb4ea222697

Running "closurecompiler:minify" (closurecompiler) task

[...]

PR #148, opened 2015-04-23
Remove extraneous quotes from output_wrapper param

alice:
@ckundo PTAL (Not sure how we missed this this whole time...)
Ref #144


ckundo:
@alice LGTM

Issue #149, opened 2015-05-01
Error evaluating site santander.com.br


davidhsv:
How to replicate: console > phantomjs audit.js http://santander.com.br
---------------------------
TypeError: undefined is not a constructor (evaluating 'op.onshow(this.op)')

  http://www.santander.com.br/document/wps/std-browser-update.js:270 in $buo
*** Begin accessibility audit results ***
An accessibility audit found 
*** End accessibility audit results ***

RickSBrown:
@davidhsv Thanks for raising this issue however it has nothing to do with the accessibility-developer-tools codebase in any way - it's actually a problem with the code in http://www.santander.com.br/document/wps/std-browser-update.js which doesn't play nice in PhantomJS.

If you wanted to raise this anywhere then the PhantomJS project would be the place, however my guess is the problem is most likely in the application JS itself, not PhantomJS.

To prove this I ran the following test:
Step 1: Create a file called `test.js` with the following code more or less copied verbatim from [here](http://phantomjs.org/quick-start.html):
```
var page = require('webpage').create();
page.open('http://santander.com.br', function(status) {
  console.log("Status: " + status);
  if(status === "success") {
    page.render('example.png');
  }
  phantom.exit();
});
```
Step 2: Run the file in phantomjs like so: `phantomjs test.js`
Step 3: Observe the exact same problem:
```
TypeError: 'undefined' is not a function (evaluating 'op.onshow(this.op)')
  http://www.santander.com.br/document/wps/std-browser-update.js:270
  http://www.santander.com.br/document/wps/std-browser-update.js:275
Status: success
```

Since the issue occurs when the accessibility-developer-tools is not loaded this test proves that it is not related to accessibility-developer-tools code. Therefore I am closing this issue.

davidhsv:
Thanks Rick for the answer !

I got futher into the problem and I found that the problem is not about the javascript error indeed, it's the iframe thats is not getting scrapped in the run method.

Please take a look at my Pull Request -> https://github.com/GoogleChrome/accessibility-developer-tools/pull/150

Regards,
David Vieira

PR #150, opened 2015-05-02
#149 Be able to evaluate page with iframes

davidhsv:
Disabling webSecurity in phantomJs -> so we can get the document root from frames.


davidhsv:
I signed it!

test/js/audit-rule-test.js
```diff
@@ -41,6 +41,16 @@
         equal(matched.length, DIV_COUNT);
     });
 
+    test("Iframe with simple DOM", function () {
+        var ifrm = document.createElement("IFRAME");
+        var container = document.getElementById('qunit-fixture');
+        container.appendChild(ifrm);
+        ifrm.contentDocument.body.appendChild(buildTestDom());
+        var matched = [];
+        axs.AuditRule.collectMatchingElements(container, matcher, matched);
```
ckundo: Is this method mutating `matched` in memory? That doesn't seem ideal. Unrelated to this PR, but since we're in the neighborhood I thought I'd point it out.

src/audits/FocusableElementNotVisibleAndNotAriaHidden.js
```diff
@@ -42,7 +42,8 @@ axs.AuditRules.addRule({
         }
         // Ignore elements which have a negative tabindex and no text content,
         // as they will be skipped by assistive technology
-        if (axs.properties.findTextAlternatives(element, {}).trim() === '')
+        var alternative = axs.properties.findTextAlternatives(element, {});
```
ckundo: This doesn't seem related to this PR.

src/js/AuditRule.js
```diff
@@ -182,6 +182,14 @@ axs.AuditRule.collectMatchingElements = function(node, matcher, collection,
             }
         }
     }
+
+    //If it is a iframe, get the contentDocument
+    if (element && element.localName == 'iframe' && element.contentDocument) {
+        axs.AuditRule.collectMatchingElements(element.contentDocument,
+                                              matcher,
+                                              collection,
+                                              opt_shadowRoot);
```
ckundo: Is `opt_shadowRoot` necessary here? Is it required for `collectMatchingElements`?

ckundo:
Check out my comment on the `opt_shadowRoot`. Aside from that question, LGTM.

ckundo:
Actually, forgot to also mention that I don't think we need to rebuild `axs_testing.js` for this to work. That change should be reverted I think.

src/audits/FocusableElementNotVisibleAndNotAriaHidden.js
davidhsv: Just a cleanup, I was getting a exception sometimes when findTextAlternatives return null (null.trim()).

src/js/AuditRule.js
davidhsv: I think so, it's a recursive method and I think it's necessary to always pass this atribute forward. For example, if we got a shadow down inside a iframe contentDocument, it will be needed.

test/js/audit-rule-test.js
davidhsv: I was justing copying the same behavior in others tests (in this same file). I don't like methods changing parameter's content too (bad smell).

davidhsv:
The axs_testing.js was generated by the grunt build. Should I revert it?

davidhsv:
Reverted file ax_testing.js to match remote master.

dist/js/axs_testing.js
```diff
@@ -1545,7 +1545,7 @@ axs.AuditRule.collectMatchingElements = function(a, b, c, d) {
         for (e = f.getDistributedNodes(), f = 0;f < e.length;f++) {
```
ckundo: @davidhsv this file needs to be checked out entirely.

src/js/AuditRule.js
ckundo: Makes sense.

test/js/audit-rule-test.js
ckundo: Yep, that's fine, thought I'd point it out as something we can improve in the future.

ckundo:
@davidhsv looks like there's a merge conflict. Once that's resolved and you remove `axs_testing.js` changes, LGTM.

alice:
I've fixed the Gruntfile so that it won't create spurious changes to `dist/axs_testing.js` any more.

src/audits/FocusableElementNotVisibleAndNotAriaHidden.js
alice: Good catch! I just submitted a separate PR for this same issue, not having seen this, sorry.

test/js/audit-rule-test.js
alice: This is intended behaviour. Since this is a recursive method, we need an accumulator to keep track of matched elements - otherwise we'd be constantly having to combine partial results.

alice:
Apologies for coming late to this. This LGTM :+1: once the merge issues are fixed.

ckundo:
@davidhsv bump. Can you resolve the merge conflicts and push again? Thanks!

davidhsv:
@ckundo Done :+1:

PR #155, opened 2015-05-20
Check for any text alternatives when assessing unlabeled images

### LINKED ISSUE #154: ImageWithoutAltText rule should not fail for images with different text alternatives
e.g. `aria-label` or `aria-labelledby`.

alice:
Fixes #154

@ckundo or @RickSBrown PTAL

### DIFF
--- a/src/audits/ImageWithoutAltText.js
+++ b/src/audits/ImageWithoutAltText.js
@@ -15,11 +15,12 @@
 goog.require('axs.AuditRules');
 goog.require('axs.browserUtils');
 goog.require('axs.constants.Severity');
+goog.require('axs.properties');
 goog.require('axs.utils');
 
 axs.AuditRules.addRule({
     name: 'imagesWithoutAltText',
-    heading: 'Images should have an alt attribute',
+    heading: 'Images should have a text alternative or presentational role',
     url: 'https://github.com/GoogleChrome/accessibility-developer-tools/wiki/Audit-Rules#ax_text_02',
     severity: axs.constants.Severity.WARNING,
     relevantElementMatcher: function(element) {
@@ -27,7 +28,15 @@ axs.AuditRules.addRule({
             !axs.utils.isElementOrAncestorHidden(element);
     },
     test: function(image) {
-        return (!image.hasAttribute('alt') && image.getAttribute('role') != 'presentation');
+        var imageIsPresentational = (image.hasAttribute('alt') && image.alt == '') || image.getAttribute('role') == 'presentation';
+        if (imageIsPresentational)
+            return false;
+        var textAlternatives = {};
+        axs.properties.findTextAlternatives(image, textAlternatives);
+        var numTextAlternatives = Object.keys(textAlternatives).length;
+        if (numTextAlternatives == 0)
+            return true;
+        return false;
     },
     code: 'AX_TEXT_02'
 });
--- a/src/js/Properties.js
+++ b/src/js/Properties.js
@@ -494,31 +494,16 @@ axs.properties.getTextFromHostLanguageAttributes = function(element,
                                                             existingComputedname,
                                                             recursive) {
     var computedName = existingComputedname;
-    if (axs.browserUtils.matchSelector(element, 'img')) {
-        if (element.hasAttribute('alt')) {
-            var altValue = {};
-            altValue.type = 'string';
-            altValue.valid = true;
-            altValue.text = element.getAttribute('alt');
-            if (computedName)
-                altValue.unused = true;
-            else
-                computedName = altValue.text;
-            textAlternatives['alt'] = altValue;
-        } else {
-            var altValue = {};
-            altValue.valid = false;
-            altValue.errorMessage = 'No alt value provided';
-            textAlternatives['alt'] = altValue;
-            var src = element.src;
-            if (typeof(src) == 'string') {
-                var parts = src.split('/');
-                var filename = parts.pop();
-                var filenameValue = { text: filename };
-                textAlternatives['filename'] = filenameValue;
-                computedName = filename;
-            }
-        }
+    if (axs.browserUtils.matchSelector(element, 'img') && element.hasAttribute('alt')) {
+        var altValue = {};
+        altValue.type = 'string';
+        altValue.valid = true;
+        altValue.text = element.getAttribute('alt');
+        if (computedName)
+            altValue.unused = true;
+        else
+            computedName = altValue.text;
+        textAlternatives['alt'] = altValue;
     }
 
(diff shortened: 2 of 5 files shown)


src/audits/ImageWithoutAltText.js
```diff
@@ -27,7 +28,16 @@ axs.AuditRules.addRule({
             !axs.utils.isElementOrAncestorHidden(element);
     },
     test: function(image) {
-        return (!image.hasAttribute('alt') && image.getAttribute('role') != 'presentation');
+        var imageIsPresentational = (image.hasAttribute('alt') && image.alt == "") || image.getAttribute('role') == 'presentation';
```
RickSBrown: total nit, question of consistency - since the strings are all single quoted should the empty string test be against `''` instead of `""`?

test/index.html
```diff
@@ -56,6 +56,7 @@
     <script src="./audits/bad-aria-attribute-value-test.js"></script>
     <script src="./audits/controls-without-label-test.js"></script>
     <script src="./audits/duplicate-id-test.js"></script>
+    <script src="./audits/image-without-alt-text-test.js"></script>
```
RickSBrown: Did you forget to add this test to source control or am I just tired?

src/audits/ImageWithoutAltText.js
```diff
@@ -27,7 +28,16 @@ axs.AuditRules.addRule({
...
+        if (imageIsPresentational)
+            return false;
+        var textAlternatives = axs.properties.getTextProperties(image);
+        if (!textAlternatives)
+            return true;
+        // filename is only present if no other option found
```
ckundo: Is this comment necessary? I'm not really sure what it means, and I'm guessing it's ok to remove it for the sake of tidiness.

src/audits/ImageWithoutAltText.js
```diff
@@ -27,7 +28,16 @@ axs.AuditRules.addRule({
...
+        if ('filename' in textAlternatives)
```
ckundo: I'm assuming this is checking if the alt text is the same as the filename.

ckundo:
Seems like this PR is missing tests.

src/audits/ImageWithoutAltText.js
alice: Fixed (nits welcome!)

test/index.html
alice: Haha, I was tired and forgot to add it. Fixed.

src/audits/ImageWithoutAltText.js
alice: It was explaining the "why" for the following line, but I've since cleaned up the logic so it's no longer necessary.

src/audits/ImageWithoutAltText.js
alice: No - the comment explains what this was doing, but the logic has changed now anyway.

alice:
@ckundo @RickSBrown PTAL!

src/audits/ImageWithoutAltText.js
```diff
@@ -27,7 +28,15 @@ axs.AuditRules.addRule({
             !axs.utils.isElementOrAncestorHidden(element);
     },
     test: function(image) {
-        return (!image.hasAttribute('alt') && image.getAttribute('role') != 'presentation');
+        var imageIsPresentational = (image.hasAttribute('alt') && image.alt == '') || image.getAttribute('role') == 'presentation';
```
ckundo: Do we care about strict equality here? I noticed you're using it below.

src/js/Properties.js
```diff
@@ -494,31 +494,16 @@ axs.properties.getTextFromHostLanguageAttributes = function(element,
...
+    if (axs.browserUtils.matchSelector(element, 'img') && element.hasAttribute('alt')) {
+        var altValue = {};
+        altValue.type = 'string';
+        altValue.valid = true;
+        altValue.text = element.getAttribute('alt');
+        if (computedName)
+            altValue.unused = true;
+        else
+            computedName = altValue.text;
+        textAlternatives['alt'] = altValue;
```
ckundo: Are we ok mixing dot notation with bracket notation for assignment? We're using both in this block of code.

ckundo:
A few style questions, no blockers though. LGTM!

src/audits/ImageWithoutAltText.js
```diff
@@ -27,7 +28,15 @@ axs.AuditRules.addRule({
...
+        if (imageIsPresentational)
+            return false;
```
RickSBrown: Question - should this be in the matcher? In other words should presentational images show a "Pass" (yes it has text alternative) or "NA" (not relevant to this test)?

test/audits/image-without-alt-text-test.js
```diff
@@ -0,0 +1,64 @@
...
+    var rule = axs.AuditRules.getRule('imagesWithoutAltText');
+
+    test('Image with no text alternative', function() {
+        var fixture = document.getElementById('qunit-fixture');
+        var img = fixture.appendChild(document.createElement('img'));
+        img.src = 'smile.jpg';
+        var expected = { elements: [img], result: axs.constants.AuditResult.FAIL };
+        deepEqual(rule.run({ scope: fixture }), expected, 'Image has no text alternative');
+    });
+
```
RickSBrown: I think two more are needed for completeness to check that presentational images are handled. One with null alt text and one with role presentation.

RickSBrown:
Likewise I see no blockers. It is a nice enhancement.
LGTM @alice

src/audits/ImageWithoutAltText.js
alice: Good question. I don't think so - I think it makes sense for the scope to be "images", and role=presentation being one technique to make the rule pass. I'll tweak the text to refer to that.

src/audits/ImageWithoutAltText.js
alice: Good point - we probably don't care here since it'll do the right thing with or without strict equality. It's not really necessary below either.

src/js/Properties.js
alice: Heh - that's an interesting question. Ideally we should use bracket notation for anything which is part of the implicit API, but since we're not doing that consistently, it doesn't really matter. Since this hasn't changed in this CL (just the indentation), I'll leave it for now, but we could consider revisiting the dot/bracket inconsistency at some point if we want to do some clean-up.

ckundo: Sounds good.

test/audits/image-without-alt-text-test.js
alice: Good idea, done.

PR #158, opened 2015-06-03
Add bower config

### LINKED ISSUE #157: Add bower support
I'm working on integrating ADT into a suite of jasmine tests loaded through `karma` and `karma-requirejs`.

Unfortunately, npm modules aren't accessible by these tests because they aren't available to the client side app. We would have to rely either on a submodule inside of our build, or manually copying `axs_testing.js` into our test build each time.

It would be useful to have a `bower` config for the project, and publish it on release.

I'm willing to work on building the config and any release automation tooling that would be needed, provided that's something you guys want.

loginx:
* Once this is merged, the module can be registered into the bower registry.
* Fixes #157

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -1,0 +1,29 @@
+{
+  "name": "accessibility-developer-tools",
+  "version": "2.7.0",
+  "homepage": "https://github.com/GoogleChrome/accessibility-developer-tools",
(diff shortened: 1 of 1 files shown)


loginx:
I signed it!

PR #160, opened 2015-06-08
Pull color code into its own file and add new suggestions algorithm.

alice:


alice:
@ckundo any objections?

src/js/AccessibilityUtils.js
```diff
@@ -13,9 +13,10 @@
 // limitations under the License.
 
 goog.require('axs.browserUtils');
+goog.require('axs.color');
+goog.require('axs.color.Color');
```
ckundo: Will this be automatically imported by the line above? Not super familiar with closure compiler.

src/js/Color.js
```diff
@@ -0,0 +1,537 @@
+// Copyright 2015 Google Inc.
```
ckundo: It there any code diff here, or is it simply moving the contents?

src/js/Properties.js
```diff
@@ -198,16 +199,28 @@ axs.properties.getContrastRatioProperties = function(element) {
...
         return null;
 
-    contrastRatioProperties['backgroundColor'] = axs.utils.colorToString(bgColor);
+    contrastRatioProperties['backgroundColor'] = axs.color.colorToString(bgColor);
     var fgColor = axs.utils.getFgColor(style, element, bgColor);
-    contrastRatioProperties['foregroundColor'] = axs.utils.colorToString(fgColor);
-    var value = axs.utils.getContrastRatioForElementWithComputedStyle(style, element);
-    if (!value)
+    contrastRatioProperties['foregroundColor'] = axs.color.colorToString(fgColor);
+    var contrast = axs.utils.getContrastRatioForElementWithComputedStyle(style, element);
```
ckundo: Nice clarification here.

src/js/Color.js
alice: I suggest reviewing the two changes in this PR separately. The first one just pulls the code into its own file; the second actually makes modifications.

test/js/color-test.js
```diff
@@ -0,0 +1,55 @@
...
+});
+test("Same color === no contrast.", function () {
+  equal(axs.color.calculateContrastRatio(this.white_, this.white_), 1);
+  equal(axs.color.calculateContrastRatio(this.black_, this.black_), 1);
+});
+test("Transparent foreground === no contrast.", function () {
+  equal(axs.color.calculateContrastRatio({"red": 0, "green": 0, "blue": 0, "alpha": 0}, this.white_), 1);
+});
+
+
```
ckundo: Is this whitespace extraneous?

test/js/color-test.js
```diff
@@ -0,0 +1,55 @@
...
+test("parses alpha values correctly", function() {
+  var colorString = 'rgba(255, 255, 255, .47)';
+  var color = axs.color.parseColor(colorString);
+  equal(color.red, 255);
+  equal(color.blue, 255);
+  equal(color.green, 255);
+  equal(color.alpha, .47);
+});
+
+module("suggestColors");
```
ckundo: Yay test coverage!

src/js/Color.js
ckundo: Got it, thanks!

src/js/AccessibilityUtils.js
alice: I don't think goog.require actually does anything useful here, to be honest, since you have to specify to the compiler what files to use anyway - it's more like documentation for the developer at this point.

Happy to take out the redundant require if you like; I just like being explicit when pulling in classes (habit from Java/C++ I guess).

ckundo: Ah, thanks for the clarity. I'm happy with whichever approach is more consistent with the style we're using in the project. Up to you.

src/js/Color.js
```diff
@@ -0,0 +1,537 @@
...
+
+    var intersection = null;
+    for (var i = 0; i < cubeFaces.length; i++) {
+        var cubeFace = cubeFaces[i];
+        intersection = axs.color.findIntersection(line, cubeFace);
+        // If intersection within [0, 1] in Z axis, it is within the cube.
+        if (intersection.z >= 0 && intersection.z <= 1)
+            break;
+    }
+    if (!intersection) {
```
ckundo: Area these guards necessary? Are there real world edge cases for these scenarios?

src/js/Color.js
```diff
@@ -0,0 +1,537 @@
...
+axs.color.toYCC = function(color) {
+    var rSRGB = color.red / 255;
+    var gSRGB = color.green / 255;
+    var bSRGB = color.blue / 255;
+
+    var r = rSRGB <= 0.03928 ? rSRGB / 12.92 : Math.pow(((rSRGB + 0.055)/1.055), 2.4);
+    var g = gSRGB <= 0.03928 ? gSRGB / 12.92 : Math.pow(((gSRGB + 0.055)/1.055), 2.4);
+    var b = bSRGB <= 0.03928 ? bSRGB / 12.92 : Math.pow(((bSRGB + 0.055)/1.055), 2.4);
+
+    return new axs.color.YCC(axs.color.multiplyMatrixVector(axs.color.YCC_MATRIX, [r, g, b]));
```
ckundo: It looks like we have a loose convention in the surrounding methods to assign a result variable and return that. Not sure if we care much, but it might be more consistent. What do you think?

ckundo:
Great work! LGTM.

test/js/color-test.js
alice: Yes :) Removed.

src/js/Color.js
alice: To be honest I don't know. I'm pretty sure it shouldn't, but I'd rather it fail non-mysteriously than mysteriously if it does. Made it an exception, since it's an exceptional case.

PR #161, opened 2015-06-08
Release automation

loginx:
This PR provides tooling for automating a lot of the `grunt` work involved in shipping out new releases.

Along with some general cleanups and enhancements (see below), the major change is the introduction of the `grunt release` task, which will perform the following tasks:

* Prompt the operator for release information (git remote, github repo info, etc...).
* Create a new build.
* Run the full suite of unit tests (currently against the source, but I'd like to make tests run against compiled bundle in the future).
* Copy the build into `dist/`.
* Bump `version` field in `package.json` and `bower.json`.
* Extract release notes from the changelog (everything above the last release header).
* Add the new release header to the changelog.
* Commit and push `package.json`, `bower.json`, `dist`.
* Create and push new git tag for the new release.
* Create a new github release against the new release tag and populate it with the release notes.

The task takes a `type` parameter, applied through the CLI as `grunt release:<type>`, where `type` is based on the `grunt-bump` plug-in.

Usage examples:

`grunt release:patch`
`grunt release:minor`
`grunt release:major`
`grunt release:2.9.5`

Minor cleanups and enhancements:

* Move temporary build artifacts to `.tmp`, and add that path to `.gitignore`.
* Normalize quotes across `Gruntfile.js` to use single quotes everywhere.
* Add `grunt clean` task to clean up `.tmp` and `dist/` automatically.
* `build` task no longer copies the compiled bundle anywhere. It's used as a low-level task now.
* New `test:unit` task wraps `qunit`.

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -1,5 +1,3 @@
-var request = require('superagent');
-
 module.exports = function(grunt) {
   'use strict';
 
@@ -7,7 +5,7 @@ module.exports = function(grunt) {
 
   grunt.initConfig({
     pkg: grunt.file.readJSON('package.json'),
-    changelog: 'CHANGELOG.md',
+    changelog: 'Changelog.md',
 
     'gh-release': {},
 
@@ -65,6 +63,14 @@ module.exports = function(grunt) {
         updateConfigs: ['pkg'],
         pushTo: "<%= grunt.config.get('gh-release.remote') %>",
         commitFiles: ['package.json', "<%= grunt.config.get('changelog') %>", 'bower.json', 'dist']
+      }
+    },
+
+    coffee: {
+      compile: {
+        files: {
+          '.tmp/util/gh_repo.js': 'src/util/gh_repo.coffee'
+        }
       }
     },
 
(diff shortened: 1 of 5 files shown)


Gruntfile.js
```diff
@@ -1,63 +1,187 @@
...
+    'gh-release': {},
 
-      },
-      'run': {}
-    },
     closurecompiler: {
       minify: {
         requiresConfig: 'git-revision',
         files: {
-          "gen/axs_testing.js": [
```
ckundo: @alice do you have an opinion on using single vs double quotes? We might want to canonize this in a styleguide.

dist/js/axs_testing.js
```diff
@@ -13,7 +13,7 @@
  * See the License for the specific language governing permissions and
```
ckundo: Do we want to be generating a new `axs_testing.js` in this PR? Seems like the PR should be limited to changes to the release process itself.

Gruntfile.js
```diff
@@ -1,63 +1,187 @@
...
+    bump: {
+      options: {
+        files: ['package.json', 'bower.json'],
+        updateConfigs: ['pkg'],
+        pushTo: "<%= grunt.config.get('gh-release.remote') %>",
+        commitFiles: ['package.json', grunt.config.get('changelog'), 'bower.json', 'dist']
+      }
+    },
+
+    prompt: {
```
ckundo: Should we use [load-grunt-config](https://www.npmjs.com/package/load-grunt-config) to break these tasks out into separate files? Maybe doesn't need to happen right away, but it's a thought.

Gruntfile.js
```diff
@@ -69,20 +193,34 @@ module.exports = function(grunt) {
     });
   });
 
+  grunt.registerTask('release', function(type) {
```
ckundo: Can we use a different argument name to be more explicit, maybe `releaseType`?

Gruntfile.js
```diff
@@ -1,63 +1,187 @@
...
 
 module.exports = function(grunt) {
+  'use strict';
+
+  require('load-grunt-tasks')(grunt);
+
   grunt.initConfig({
-    'git-describe': {
-      options: {
+    pkg: grunt.file.readJSON('package.json'),
```
ckundo: Is this required for the `bump` task? I'm not seeing any new configuration in `package.json`, so wanted to check if this is necessary.

Gruntfile.js
```diff
@@ -1,63 +1,187 @@
...
+      options: {
+        files: ['package.json', 'bower.json'],
+        updateConfigs: ['pkg'],
+        pushTo: "<%= grunt.config.get('gh-release.remote') %>",
+        commitFiles: ['package.json', grunt.config.get('changelog'), 'bower.json', 'dist']
+      }
+    },
+
+    prompt: {
+      'gh-release': {
```
ckundo: I *think* github automatically creates releases if you use semantic versioning tags (`v0.0.1`). I'm looking for documentation on that, but I've observed it happening on my projects. If that's the case, we could avoid some of these steps.

ckundo: I'm mistaken, github shows all tags in the releases tabs *only if* there are explicit releases defined.

Gruntfile.js
```diff
@@ -1,63 +1,187 @@
...
+        options: {
+          questions: [
+            {
+              config: 'gh-release.remote',
+              type: 'input',
+              message: 'Git Remote (usually GoogleChrome or origin)',
```
ckundo: Should we put the default value in the message? `message: 'Git Remote (defaults to GoogleChrome)',

Gruntfile.js
```diff
@@ -1,63 +1,187 @@
...
+    grunt.log.ok("Changelog updated, and release notes extracted.");
+  });
+
+  grunt.registerTask('gh-release', function() {
+    var config = grunt.config.get('gh-release');
+    var pkg = grunt.config.get('pkg');
+    var done = this.async();
+
+    request
+      .post('https://api.github.com/repos/' + config.repo + '/releases')
```
ckundo: Should these urls be defined as config values, instead of inline?

ckundo:
Looks good, I'll defer to @alice on final thumbsup.

dist/js/axs_testing.js
loginx: Yeah agreed, that change is a byproduct of the fact that there was an initial change that involved nuking dist and adding it to gitignore, and we no longer want that, so I had to re-add it.

I'll just revert that file to its original version.

Gruntfile.js
loginx: I don't think the task configs themselves are consuming too much space, but it would definitely make sense to move custom tasks away from this file.

Gruntfile.js
loginx: It's required for a lot of the custom tasks that need to access the `version` field from that file.

Gruntfile.js
loginx: No, the `grunt-prompt` task does that automatically for us.

Gruntfile.js
loginx: The only scenario I can think of where you'd query the github API with a different base URL would be if you deployed an internal github enterprise system. If that's in the cards, we can certainly set this up as a config option.

That specific grunt task is being refactored in my experimental branch though, after discussions with @alice regarding how she'd like github releases to be managed.

Gruntfile.js
alice: I'm not too fussed at this stage - the consistency in the new version is nice.

I'm not sure what tooling we have that integrates with GitHub, but I can look into it - would be nice to have some kind of hook to check this stuff rather than having to go back and forth in reviews.

alice:
:+1: Go ahead and pull once the merge conflict issue is resolved.

PR #162, opened 2015-06-09
Export AMD module

loginx:
* Wrap the output into an AMD compatible module if possible. Otherwise, keep exporting `axs` and `goog` into the global scope.
* Fresh rebuild of `axs_testing.js` bundle.
* Update `bower.json` config to advertise support for AMD module.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,3 +1,7 @@
+### Enhancements
+
+* Implement support for AMD loaders.
+
 ### New rules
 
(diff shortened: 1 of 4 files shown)


ckundo:
Maybe it makes sense to write an integration test that requires `axs_testing` so we can get test coverage on this. Otherwise, LGTM!

ckundo:
@loginx do you want to merge this (once the conflicts are resolved)?

loginx:
Will do. I'll have a look at the conflicts in a bit. They probably just stem from the fact that there's been some new builds since this was proposed. Shouldn't be too difficult to merge.

PR #163, opened 2015-06-09
Use SPDX License IDs

loginx:
* Both NPM and Bower specify that the `license` field should be a valid
SPDX license ID.
  * https://github.com/bower/bower.json-spec#license
  * https://docs.npmjs.com/files/package.json#license

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -15,7 +15,7 @@
     "testing",
     "WCAG"
   ],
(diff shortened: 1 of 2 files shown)


ckundo:
Deferring to @alice in case there are Google specific guidelines around this.

alice:
LGTM

PR #165, opened 2015-06-10
Add support for config payload for AuditConfiguration

### LINKED ISSUE #164: Enhancement: Provide a config object to AuditConfiguration constructor
I think it would be nice to be able to provide a configuration payload to the `AuditConfiguration` constructor instead of creating it and then setting each option one by one.

This would allow for config payloads to be reusable when running a batch of audits.

i.e.

```javascript
var config, auditConfig, audit, report;

config = {
  scope: document.querySelector('.active'),
  showUnsupportedRulesWarning: false,
  maxResults: 1
};

auditConfig = new axs.AuditConfiguration(config);
audit = axs.Audit.run(auditConfig);
report = axs.Audit.createReport(audit);
```

loginx:
* Overwrite `AuditConfiguration` properties defined in the constructor with properties found in payload parameter.
* Add test for new behaviour.
* Add usage example in docs.
* Fixes #164

### DIFF
--- a/src/js/Audit.js
+++ b/src/js/Audit.js
@@ -20,9 +20,22 @@ goog.provide('axs.Audit');
 goog.provide('axs.AuditConfiguration');
 
 /**
+ * @typedef {{
+ *   scope: (Element|undefined),
+ *   auditRulesToRun: (Array.<String>|undefined),
+ *   auditRulesToIgnore: (Array.<String>|undefined),
+ *   maxResults: (number|undefined),
+ *   withConsoleApi: (boolean|undefined),
+ *   showUnsupportedRulesWarning: (boolean|undefined)
+ * }}
+ */
+axs.AuditConfigurationOptions_;
+
+
+/**
  * Object to hold configuration for an Audit run.
  * @constructor
- * @param {?Object=} config Configuration object
+ * @param {?axs.AuditConfigurationOptions_=} config Configuration object
  */
 axs.AuditConfiguration = function(config) {
     if (config == null) {
--- a/src/js/Audit.js
+++ b/src/js/Audit.js
@@ -22,8 +22,13 @@ goog.provide('axs.AuditConfiguration');
 /**
  * Object to hold configuration for an Audit run.
  * @constructor
+ * @param {?Object=} config Configuration object
  */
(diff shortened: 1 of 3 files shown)


src/js/Audit.js
```diff
@@ -22,8 +22,13 @@ goog.provide('axs.AuditConfiguration');
 /**
  * Object to hold configuration for an Audit run.
  * @constructor
+ * @param {?Object=} config Configuration object
  */
-axs.AuditConfiguration = function() {
+axs.AuditConfiguration = function(config) {
```
ckundo: Cool! Me gusta.

ckundo:
This is an exciting change, great work! LGTM

src/js/Audit.js
```diff
@@ -78,6 +83,12 @@ axs.AuditConfiguration = function() {
      */
     this.showUnsupportedRulesWarning = true;
 
+    for (var prop in this) {
+        if ((this.hasOwnProperty(prop)) && (prop in config)) {
+            this[prop] = config[prop];
```
alice: This is cool, but is not going to work for `severity`, `ignoreSelectors` or `ruleConfig`. We'll probably need to special case those.

loginx:
@alice That's correct. That can be done but we have to determine a reasonable format for the data to pass to these methods in the config payload. 

It would probably be a good idea to merge this in first, and open an issue for automating these methods through config as well.

src/js/Audit.js
```diff
@@ -22,8 +22,13 @@ goog.provide('axs.AuditConfiguration');
 /**
  * Object to hold configuration for an Audit run.
  * @constructor
+ * @param {?Object=} config Configuration object
```
alice: Can you add more detailed documentation for this, stating what can and cannot be configured this way?

loginx: Will do.

ckundo:
Looking forward to getting this merged! @loginx do you need an support on this?

loginx:
@ckundo I completely forgot about this. I thought it had been merged long ago. I'll have a look at this tonight as well.

alice:
:+1: go ahead and pull

PR #169, opened 2015-06-15
Move gh_repo.coffee to pre-existing scripts/ directory

alice:
@loginx please take a look - missed this earlier, but we have scripts/ for sources which don't get built into the lib.


loginx:
:+1:

PR #170, opened 2015-06-15
Enhance dry-run release tasks

loginx:
* Add `—dry-run` option support for the `changelog` and `gh-release` tasks.
* Skip the bundle compilation, unit testing and bundle distribution tasks when `—dry-run` is specified.

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -121,6 +121,7 @@ module.exports = function(grunt) {
   grunt.registerTask('changelog', function(type) {
     grunt.task.requires('bump-only:' + type);
 
+    var dryRun = grunt.option('dry-run');
     var config = {
       data: {
         version: grunt.config.get('pkg.version'),
@@ -136,6 +137,7 @@ module.exports = function(grunt) {
     var headerTpl = "## <%= version %> - <%= releaseDate %>\n\n";
     var header = grunt.template.process(headerTpl, config);
 
+    grunt.log.ok('changelog: Extracting release notes.');
     if (contents.length > 0) {
       if ((stopIndex = contents.search(stopRegex)) !== -1) {
         releaseNotes = contents.slice(0, stopIndex);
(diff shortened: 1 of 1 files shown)


Gruntfile.js
```diff
@@ -144,8 +146,12 @@ module.exports = function(grunt) {
 
     grunt.config.set("gh-release.release-notes", releaseNotes);
 
-    grunt.file.write(dest, "" + header + contents);
-    grunt.log.ok("Changelog updated, and release notes extracted.");
+    if (dryRun) {
+      grunt.log.ok('changelog (dry): Prepending header to ' + dest);
```
alice: Can we log the release notes and header here as well?

Gruntfile.js
```diff
@@ -144,8 +146,12 @@ module.exports = function(grunt) {
...
+    } else {
+      grunt.file.write(dest, "" + header + contents);
+    }
+    grunt.log.ok('changelog: ' + dest + ' updated, and release notes extracted.');
```
alice: Maybe this should go inside the previous else block.

Gruntfile.js
```diff
@@ -216,17 +233,25 @@ module.exports = function(grunt) {
...
-      'prompt:gh-release',
-      'build',
-      'test:unit',
-      'copy:dist',
+    var dryRun = grunt.option('dry-run');
+
+    var tasks = ['prompt:gh-release'];
+
+    if (dryRun) {
+      grunt.log.ok('Skipping build, test:unit and copy:dist tasks in dry-run');
```
alice: We might want to keep build and test in there, since they don't modify any files actually in git? What do you think?

Gruntfile.js
loginx: Not really because it's really a notification that the task has completed. We can certainly reword it though.

Gruntfile.js
loginx: Sure, but logging the release notes to the console might generate quite a bit of output if you have a large release.

Gruntfile.js
loginx: Yeah I thought about it, but I figured that the point of the dry-run for the `release` task would always be to see what tags and releases would get generated. Those tasks would just slow things down in that context.

I have no problem bringing them back in.

Gruntfile.js
alice: In that case, I think it's fine to update Changelog but not commit it - that way, I can just git checkout -- . once I'm done eyeballing what the script spits out.

alice:
:+1:

alice:
Not sure why this has merge conflicts :frowning:

ckundo:
@alice merge conflicts are resolved here, do you want to merge?

alice:
@loginx can merge this any time, he has write access. Otherwise, I'm happy to do it tomorrow (assuming he's busy with other stuff)

loginx:
I just returned from vacation. I'll take care of this today.

PR #173, opened 2015-06-22
Fix hidden case for FocusableElementNotVisibleNotAriaHidden

alice:
@ckundo PTAL

### DIFF
--- a/src/audits/FocusableElementNotVisibleAndNotAriaHidden.js
+++ b/src/audits/FocusableElementNotVisibleAndNotAriaHidden.js
@@ -42,7 +42,8 @@ axs.AuditRules.addRule({
         }
         // Ignore elements which have a negative tabindex and no text content,
         // as they will be skipped by assistive technology
-        if (axs.properties.findTextAlternatives(element, {}).trim() === '')
+        var textAlternatives = axs.properties.findTextAlternatives(element, {});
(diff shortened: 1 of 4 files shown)


src/audits/FocusableElementNotVisibleAndNotAriaHidden.js
```diff
@@ -42,7 +42,8 @@ axs.AuditRules.addRule({
         }
         // Ignore elements which have a negative tabindex and no text content,
         // as they will be skipped by assistive technology
-        if (axs.properties.findTextAlternatives(element, {}).trim() === '')
+        var textAlternatives = axs.properties.findTextAlternatives(element, {});
+        if (!textAlternatives || textAlternatives.trim() === '')
```
ckundo: I think we should add `===` as a best practice to the yet-to-be styleguide.

ckundo:
LGTM! Was the original test passing because it was not running in the browser?

alice:
The original test wasn't running at all because it wasn't included in the non-browser tests.

src/audits/FocusableElementNotVisibleAndNotAriaHidden.js
alice: Sounds good :smile:

PR #177, opened 2015-06-29
url update to prevent MIME error

georgebatalinski:
Refused to execute script from 'https://raw.github.com/GoogleChrome/accessibility-developer-tools/stable/dist/js/axs_testing.js' because its MIME type ('text/plain') is not executable, and strict MIME type checking is enabled.

This is to prevent this error 
see http://stackoverflow.com/questions/17341122/link-and-execute-external-javascript-file-hosted-on-github

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,7 +13,7 @@ There is also a collection of accessibility-related utility code, including but 
 
 To include just the javascript rules, require the following file:
 
-    https://rawgit.com/GoogleChrome/accessibility-developer-tools/stable/dist/js/axs_testing.js
+    https://cdn.rawgit.com/GoogleChrome/accessibility-developer-tools/master/dist/js/axs_testing.js
 
   `git 1.6.5` or later: 
 
--- a/README.md
+++ b/README.md
@@ -13,7 +13,7 @@ There is also a collection of accessibility-related utility code, including but 
 
 To include just the javascript rules, require the following file:
 
-    https://raw.github.com/GoogleChrome/accessibility-developer-tools/stable/dist/js/axs_testing.js
+    https://rawgit.com/GoogleChrome/accessibility-developer-tools/stable/dist/js/axs_testing.js
 
   `git 1.6.5` or later: 
 


georgebatalinski:
I signed it!

README.md
```diff
@@ -13,7 +13,7 @@ There is also a collection of accessibility-related utility code, including but
 
 To include just the javascript rules, require the following file:
 
-    https://raw.github.com/GoogleChrome/accessibility-developer-tools/stable/dist/js/axs_testing.js
+    https://rawgit.com/GoogleChrome/accessibility-developer-tools/stable/dist/js/axs_testing.js
```
loginx: Good catch on pointing to the production URL. According to rawgit.com though, the URL we should use is https://cdn.rawgit.com/GoogleChrome/accessibility-developer-tools/master/dist/js/axs_testing.js

Could you please update this PR with that change?

georgebatalinski: Got it

README.md
```diff
@@ -13,7 +13,7 @@ There is also a collection of accessibility-related utility code, including but
 
 To include just the javascript rules, require the following file:
 
-    https://raw.github.com/GoogleChrome/accessibility-developer-tools/stable/dist/js/axs_testing.js
+    https://cdn.rawgit.com/GoogleChrome/accessibility-developer-tools/master/dist/js/axs_testing.js
```
ckundo: This content would be hosted by a third-party that is unaffiliated with Github[1]. That makes me uncomfortable and would require vetting the third-party. @alice what do you think?

[1]https://stackoverflow.com/questions/17341122/link-and-execute-external-javascript-file-hosted-on-github#comment30352321_18049842

ckundo:
@alice does Google have a CDN we could release to? That would resolve this issue without hosting on an unvetted third party.

alice:
Yeah, need to think about this. I'll look into whether rawgit is ok, as well.

alice:
Yeah, can't use rawgit.com unfortunately. Working on an alternative.

PR #186, opened 2015-07-15
Add new audit rule that warns about bad page titles

kristapsmelderis:


kristapsmelderis:
"Open the pod bay doors, googlebot"

Seriously though, what is the problem,
I went through the arduous steps to completely delete my work and re-create it with new git global config for my e-mail and...no cake still.

PR #187, opened 2015-07-15
Add new audit rule that warns about bad page title

kristapsmelderis:


src/audits/PageWithBadTitle.js
```diff
@@ -0,0 +1,43 @@
...
+    name: 'pageWithBadTitle',
+    heading: 'The web page should have a title less than 66 characters long and does not contain bad characters',
+    url: 'https://github.com/GoogleChrome/accessibility-developer-tools/wiki/Audit-Rules#TODO-WIKIENTRY',
+    severity: axs.constants.Severity.WARNING,
+    relevantElementMatcher: function(element) {
+        return element.tagName.toLowerCase() == "title";
+    },
+    test: function(scope) {
+        // test length of title
+        if (document.title.length >= 66)
```
alice: `document` may not be what you expect here.

`scope` is an element which `relevantElementMatcher` returns true on (see https://github.com/GoogleChrome/accessibility-developer-tools/blob/master/src/js/AuditRule.js#L243) and may be inside an iframe, in which case its owner document will be something other than `document` which will be the owner document of the context where this script is running. For example, if I inject this script into my page, `document` will be the top level document for my page, but `scope` may be inside an iframe which has its own `<title>`, which would be what I would want to check.

Since the element being passed in will be a `<title>`, you may want to rename `scope` to `title` or `titleEl` (like how https://github.com/GoogleChrome/accessibility-developer-tools/blob/master/src/audits/ImageWithoutAltText.js calls the argument to its `test()` function `image`). Then you can check the `textContent` of the title element.

src/audits/PageWithBadTitle.js
```diff
@@ -0,0 +1,43 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+goog.require('axs.AuditRules');
+goog.require('axs.constants.Severity');
+
+axs.AuditRules.addRule({
+    name: 'pageWithBadTitle',
+    heading: 'The web page should have a title less than 66 characters long and does not contain bad characters',
```
alice: Can you explain the reasoning behind this rule?

kristapsmelderis: This suggested rule is to help users that depend on screen readers to browse the web;
Comes from a list of minor issues, collected by an accessibility expert who uses a screen reader on a daily basis;
I do not have any demographics research done on such a thing though;
I left the severity to a warning.

src/audits/PageWithBadTitle.js
kristapsmelderis: I used https://github.com/GoogleChrome/accessibility-developer-tools/blob/master/src/audits/PageWithoutTitle.js
as a reference material.
I will try to fix the script as you suggested.

src/audits/PageWithBadTitle.js
ckundo: @kristapsmelderis can you be more specific about the genesis of this rule? Is there a corresponding standard that drove this out? Why 66 characters?

I'm hesistant to merge this unless there's a convention or standard backing it up.

src/audits/PageWithBadTitle.js
```diff
@@ -0,0 +1,32 @@
...
+    url: 'https://github.com/GoogleChrome/accessibility-developer-tools/wiki/Audit-Rules#TODO-WIKIENTRY',
+    severity: axs.constants.Severity.WARNING,
+    relevantElementMatcher: function(element) {
+        return element.tagName.toLowerCase() == "title";
+    },
+    test: function(title) {
+        var titleText = title.textContent;
+        var badChars = /([-/\\]|[.]$)/;
+        return titleText.length > 65 || badChars.test(titleText);
```
ckundo: Why this length and not something else?

test/audits/page-with-bad-title-test.js
```diff
@@ -0,0 +1,65 @@
+module("Page titles");
+
+test("Page title must be less than 66 characters in length", function() {
+    var rule = axs.AuditRules.getRule('pageWithBadTitle');
+
+    // Save the existing title for later use
```
ckundo: I think the code here is self-explanatory, so these comments aren't necessary.

ckundo:
Thanks for the PR @kristapsmelderis. I think there is a big open question here about whether there is any corresponding standard or widely accepted convention associated with this rule. I understand the *spirit* of the rule, but the specifics seem a little arbitrary. Can you fill us in more?

kristapsmelderis:
Hello @ckundo 

Apologies, it has been a while;
The main reason for the rule was from practical experience using a screen reader.
- A long page title is hard to keep in memory;
- Characters that are literally read out by screen readers ( e.g. '.' as "dot" or "full stop") distract the person and the context can be lost;

As for standards. Nothing concrete, however:
https://www.w3.org/Provider/Style/TITLE.html

"The title should ideally be less than 64 characters in length." - the focus is on technical aspects. Nothing about accessibility.

The examples show presence of symbols such as '.' - I can only assume that the use case: "what if titles are read by screen readers" has been left out from standard considerations when they were written.

Issue #183, opened 2015-07-15
TypeError: null is not an object (evaluating 'axs.properties.findTextAlternatives(a, {}).trim')


kennyw12:
There is a strange error that is thrown when running the CLI against a page that contains:

```html
<!-- Button trigger modal -->
<button class="btn btn-primary" data-toggle="modal" data-target="#myModal">
  Launch modal
</button>

<!-- Modal Basic -->
<div class="modal fade modal-basic" id="myModal" tabindex="-1" role="dialog" aria-labelledby="myModalLabel" aria-hidden="true">
  <div class="modal-dialog">
    <div class="modal-content">
      <div class="modal-header">
        <button type="button" class="close" data-dismiss="modal"><span aria-hidden="true">&times;</span><span class="sr-only">Close</span></button>
        <h4 class="modal-title" id="myModalLabel">Modal title</h4>
      </div>
      <div class="modal-body">
        ...
      </div>
      <div class="modal-footer">
        <button type="button" class="btn btn-lowlight" data-dismiss="modal">Close</button>
        <button type="button" class="btn btn-primary">Save changes</button>
      </div>
    </div>
  </div>
</div>
```

The error that is thrown is:

```
TypeError: null is not an object (evaluating 'axs.properties.findTextAlternatives(a, {}).trim')

  ../../dist/js/axs_testing.js:1868 in relevantElementMatcher
  ../../dist/js/axs_testing.js:1530 in collectMatchingElements
  ../../dist/js/axs_testing.js:1553 in collectMatchingElements
  ../../dist/js/axs_testing.js:1553 in collectMatchingElements
  ../../dist/js/axs_testing.js:1553 in collectMatchingElements
  ../../dist/js/axs_testing.js:1553 in collectMatchingElements
  ../../dist/js/axs_testing.js:1553 in collectMatchingElements
  ../../dist/js/axs_testing.js:1560 in run
  ../../dist/js/axs_testing.js:1708 in run
  phantomjs://webpage.evaluate():2
  phantomjs://webpage.evaluate():4
null
```

We've tried poking around at the source of this issue by removing and changing the `tabindex` property in `.modal-basic`. Removing `tabindex`, or changing it to a positive value allows CLI to run properly.

cc/ @gpleiss, @matt-royal, @stubbornella, @ctaymor

geekdave:
I'm seeing this as well.  Reproducible easily using:

```
$ a11y http://www.walmart.com --verbose
[...]

PR #188, opened 2015-07-22
Fixes issue #183

RickSBrown:
Actually this is a workaround, not a fix because the bug is in closure compiler. I'll break it down in detail below.

First understand that the function `axs.properties.findTextAlternatives` can legitimately return `null`.
Then take a look at the way we call it:

```
var textAlternatives = axs.properties.findTextAlternatives(element, {});
if (!textAlternatives || textAlternatives.trim() === '')
```

All good. But then take a look at what closure compiler produces:
```
return "" === axs.properties.findTextAlternatives(a, {}).trim() ? !1 : !0;
```
That will throw an exception if the return value is null.
The easiest workaround is to get closure compiler to produce something else. The proposed change does not change the logic of the audit but results in safe code generated by closure compiler:
```
a = axs.properties.findTextAlternatives(a, {});
  return null == a || "" === a.trim() ? !1 : !0;
```

### DIFF
--- a/src/audits/FocusableElementNotVisibleAndNotAriaHidden.js
+++ b/src/audits/FocusableElementNotVisibleAndNotAriaHidden.js
@@ -43,7 +43,7 @@ axs.AuditRules.addRule({
         // Ignore elements which have a negative tabindex and no text content,
         // as they will be skipped by assistive technology
         var textAlternatives = axs.properties.findTextAlternatives(element, {});
-        if (!textAlternatives || textAlternatives.trim() === '')
+        if (textAlternatives == null || textAlternatives.trim() === '')
             return false;
 
         return true;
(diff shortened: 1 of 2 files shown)


RickSBrown:
@alice PTAL

src/audits/FocusableElementNotVisibleAndNotAriaHidden.js
```diff
@@ -43,7 +43,7 @@ axs.AuditRules.addRule({
         // Ignore elements which have a negative tabindex and no text content,
         // as they will be skipped by assistive technology
         var textAlternatives = axs.properties.findTextAlternatives(element, {});
-        if (!textAlternatives || textAlternatives.trim() === '')
+        if (textAlternatives == null || textAlternatives.trim() === '')
```
alice: Let's use === null - closure compile should catch if `findTextAlternatives` weirdly returns undefined anywhere.

alice:
Please update the Changelog: https://github.com/GoogleChrome/accessibility-developer-tools/blob/master/Changelog.md - manual process to help me write release notes. Sorry for the overhead :(

Once that's done and the nit addressed, this is :+1: - definitely an improvement regardless of the closure weirdness.

RickSBrown:
No worries about the changelog overhead, it is well worth the effort IMHO.

PR #189, opened 2015-07-23
Add .matches() as default choice for browserUtils.matchSelector

jdan:
Formerly known as matchesSelector(), with many vendor prefixes as seen in this function.
I'm using ADT in a jsdom environment, which doesn't support any vendor-prefixed variations
of this function.

http://caniuse.com/#feat=matchesselector

### DIFF
--- a/src/js/BrowserUtils.js
+++ b/src/js/BrowserUtils.js
@@ -15,6 +15,7 @@
 goog.provide('axs.browserUtils');
 
 /**
+ * Use Webkit matcher when matches() is not supported.
  * Use Firefox matcher when Webkit is not supported.
  * Use IE matcher when neither webkit nor Firefox supported.
  * @param {Element} element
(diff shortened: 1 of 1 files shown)


src/js/BrowserUtils.js
```diff
@@ -15,6 +15,7 @@
 goog.provide('axs.browserUtils');
 
 /**
+ * Use Webkit matcher when matches() is not supported.
```
ckundo: Should this read "Use Webkit matcher when matches() *is* supported."?

ckundo:
It'd be nice to find a way to unit test this. LGTM otherwise.

src/js/BrowserUtils.js
ckundo: Ah, I see you were just adding missing documentation, this makes sense to me now.

jdan: Yeah, tried to follow the pattern we had going on here

jdan:
Thanks!

jdan:
@ckundo @alice FWIW there was no change to the changelog to accompany this

alice:
Ah, thanks for the heads up. Mind doing a follow up with the Changelog change? We'll get used to this flow eventually...

ckundo:
Sorry, my mistake! Now I know.

PR #195, opened 2015-07-25
Fixes #178 - Unsupported ARIA attribute fails with global attributes

### LINKED ISSUE #178: UnsupportedAriaAttribute audit test fails when input has aria-label
The following HTML results in AX_ARIA_10 rule (unsupported aria attribute) failing.  This attribute should be valid.

```javascript
<input aria-label="my test" type="text">
```

This is the first time I've looked at these tools so I am not very familiar with the project, but the `attributeName` check in the following line of code looks suspect to me:

```javascript
if (axs.constants.ARIA_PROPERTIES.hasOwnProperty(lookupName) && !(attributeName in supported)) {
    return true;
}
```

RickSBrown:
```supported``` was sometimes an object, sometimes an array.
Probably best as an object for fast lookup.

### CODE REVIEW

**Comment by RickSBrown** [comment]
@alice PTAL

File: src/audits/UnsupportedAriaAttribute.js [hunk 1]
```diff
@@ -49,7 +51,15 @@ goog.require('axs.utils');
             }
             else {
                 // This test ignores the fact that some HTML elements should not take even global attributes.
-                supported = axs.constants.GLOBAL_PROPERTIES;
+                if(globalProperties) {  // if we have previously computed this value, use it
```
alice: This is nice, but I think it's an unnecessary optimisation. Other initialisation happens right inside Constants.js, e.g. https://github.com/GoogleChrome/accessibility-developer-tools/blob/master/src/js/Constants.js#L982 - could you use this pattern instead?

test/audits/unsupported-aria-attribute-test.js
```diff
@@ -125,4 +125,12 @@
         deepEqual(rule.run({ scope: fixture }), expected);
     });
 
+	test('Element with no role and global aria- attributes', function() {
```
alice: Nice test :)

RickSBrown:
@alice hopefully it's a bit more consistent with other constants now... Happy to tweak it further if required.

RickSBrown:
@googlebot alice is the author of those commits and has signed the CLA / works for Google.

RickSBrown:
alice (the project owner) is the author of the other commits and has signed
the CLA / works for Google

alice:
:+1:

Issue #182, opened 2015-07-11
if aria-value/now/min/max is a decimal value, warning "ARIA state and property values must be valid" 


akhoury:
According to the spec, a decimal value is fine. What am I missing?

http://www.w3.org/TR/wai-aria/states_and_properties#aria-valuenow

> The value of aria-valuenow is a decimal number. If the range is a set of numeric values, then aria-valuenow is one of those values. For example, if the range is [0, 1], a valid aria-valuenow is 0.5. A value outside the range, such as -2.5 or 1.1, is invalid.

### `aria-valuenow` and `aria-valuemax` warnings
![screen shot 2015-07-11 at 12 40 00 pm](https://cloud.githubusercontent.com/assets/1398375/8634523/1325739e-27ca-11e5-84a9-12a3c7d1aa0a.png)

### `aria-valuemin` warning
![screen shot 2015-07-11 at 12 33 41 pm](https://cloud.githubusercontent.com/assets/1398375/8634517/8db89114-27c9-11e5-97e6-dd6a3d78fb41.png)

Thank you!

akhoury:
is there a reason why both `integer` and 'decimal` values are validated the same way?

https://github.com/GoogleChrome/accessibility-developer-tools/blob/48b283ed84ab8544095b31c4b2052d523f9ffd8a/src/js/AccessibilityUtils.js#L692-L707

```javascript
    case "integer":
    case "decimal":
        var validNumber = axs.utils.isValidNumber(value);
        if (!validNumber.valid) {
            result.valid = false;
            result.reason = validNumber.reason;
            return result;
        }
        if (Math.floor(validNumber.value) != validNumber.value) { // <--- this is testing decimals as well
            result.valid = false;
            result.reason = '' + value + ' is not a whole integer';
        } else {
            result.valid = true;
            result.value = validNumber.value;
        }
        return result;
```

Looks like an easy fix

RickSBrown:
Looks like the intention was to fall through from `number` to `decimal` not `integer` to `decimal`.

RickSBrown:
PR #196 will fix this.

RickSBrown:
Closed by #196

PR #196, opened 2015-07-25
Fixes issue #182 and adds a unit test to check this condition.

RickSBrown:
Looks like the intent was for `decimal` and `number` to be validated the same way instead of `decimal` and `integer`.

### DIFF
--- a/src/js/AccessibilityUtils.js
+++ b/src/js/AccessibilityUtils.js
@@ -654,7 +654,6 @@ axs.utils.getRoles = function(element, implicit) {
             roleObject.valid = result.valid = true;
         } else {
             roleObject.valid = false;
-
         }
         result.roles.push(roleObject);
     }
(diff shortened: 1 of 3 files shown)


RickSBrown:
@alice PTAL

src/js/AccessibilityUtils.js
```diff
@@ -654,7 +654,7 @@ axs.utils.getRoles = function(element, implicit) {
             roleObject.valid = result.valid = true;
         } else {
             roleObject.valid = false;
-            
+
```
alice: Weird that there was a bunch of spaces here before; can we remove the empty line as well?

src/js/AccessibilityUtils.js
```diff
@@ -720,11 +719,13 @@ axs.utils.getAriaPropertyValue = function(propertyName, value, element) {
...
     case "number":
         var validNumber = axs.utils.isValidNumber(value);
-        if (validNumber.valid) {
-            result.valid = true;
-            result.value = validNumber.value;
+        if (!validNumber.valid) {
+            result.valid = false;
+            result.reason = validNumber.reason;
+            return result;
         }
```
alice: It also seems weird that we don't have a return for the valid case here (undoubtedly my fault) - mind adding one?

alice:
:+1: once comments addressed. Nice work!

src/js/AccessibilityUtils.js
RickSBrown: Hehe - I was scared to touch it - I'll fix it up :)

PR #197, opened 2015-07-25
Add a special case to handle color `"transparent"` to fix #180.

### LINKED ISSUE #180: axs.utils.getBgColor returns null for unstyled elements in Firefox
Hey @alice,

The following code segment:

```js
axs.utils.getBgColor = function(style, element) {
    var bgColorString = style.backgroundColor;
    var bgColor = axs.color.parseColor(bgColorString);
    if (!bgColor) 
        return null;
    ...
};
```

Will return `null` if `bgColor` is falsy. Looking at `axs.color.parseColor`:

```js
axs.color.parseColor = function(colorString) {
    var rgbRegex = /^rgb\((\d+), (\d+), (\d+)\)$/;
    var match = colorString.match(rgbRegex);
    if (match) {
        var r = parseInt(match[1], 10);
        var g = parseInt(match[2], 10);
        var b = parseInt(match[3], 10);
        var a = 1;
        return new axs.color.Color(r, g, b, a);
    }
    var rgbaRegex = /^rgba\((\d+), (\d+), (\d+), (\d*(\.\d+)?)\)/;
    match = colorString.match(rgbaRegex);
    if (match) {
        var r = parseInt(match[1], 10);
        var g = parseInt(match[2], 10);
        var b = parseInt(match[3], 10);
        var a = parseFloat(match[4]);
        return new axs.color.Color(r, g, b, a);
    }
    return null;
};
```

We see that we expect `colorString` to be valued at either `rgb(...)` or `rgba(...)`. However, in FF, elements will have a default `backgroundColor` of `"transparent"`.

This causes `parseColor` (and consequently, `getBgColor`) to return `null`, eventually causing `getContrastRatioForElement` to return `null`.

You can [see the discrepancy in this jsbin](https://output.jsbin.com/haluta/2) by running it in both Chrome and FF.

RickSBrown:
Fixes issue #180 
This bug is responsible for the only failing unit test (`LowContrast - "Opacity is handled"`) in:
* Firefox
* IE11
* MS Edge

This fix makes 100% of unit tests pass in those browsers.

Also checked Chrome, Opera and Safari, all 100%.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -9,6 +9,7 @@
 ### Bug fixes:
 
 * Work around null pointer exception caused by closure compiler issue (#183).
+* Add a special case to handle color `"transparent"` to fix #180.
 
 ## 2.8.0 - 2015-07-24
 
--- a/src/js/Color.js
+++ b/src/js/Color.js
@@ -137,6 +137,9 @@ axs.color.luminanceRatio = function(luminance1, luminance2) {
  * @return {?axs.color.Color}
  */
 axs.color.parseColor = function(colorString) {
+    if(colorString === "transparent") {
+        return new axs.color.Color(0, 0, 0, 0);
+    }
     var rgbRegex = /^rgb\((\d+), (\d+), (\d+)\)$/;
     var match = colorString.match(rgbRegex);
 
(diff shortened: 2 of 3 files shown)


RickSBrown:
@alice PTAL

jdan:
Thanks @RickSBrown!

alice:
This is :+1: but could you possibly add a trivial unit test passing in "transparent" and asserting that return value, with a comment pointing to #180, just to make sure we don't accidentally regress this in some later "cleanup"?

PR #200, opened 2015-07-26
Fixes #199 by making axs.constants.GLOBAL_PROPERTIES the set

### LINKED ISSUE #199: UnsupportedAriaRoles has false positives on global attributes
e.g. `<div aria-hidden="true">` is returned as an error.

alice:
...rather than the list.

@RickSBrown PTAL


src/js/Properties.js
```diff
@@ -703,8 +703,7 @@ axs.properties.getAriaProperties = function(element) {
  */
 axs.properties.getGlobalAriaProperties = function(element) {
     var globalProperties = {};
-    for (var i = 0; i < axs.constants.GLOBAL_PROPERTIES.length; i++) {
-        var property = axs.constants.GLOBAL_PROPERTIES[i];
+    for (var property in axs.constants.GLOBAL_PROPERTIES) {
```
RickSBrown: i'd consider adding ` && axs.constants.GLOBAL_PROPERTIES.hasOwnProperty(property)` but it's not truly necessary here

RickSBrown:
@alice LGTM
This is great, i totally missed `propertiesSet` in PR #195, * facepalm * - sorry.

I think this bug is a duplicate of #178 so this PR should fix both and we can just close #195.

alice:
Oh man I'm so :flushed: that I forgot about your PR and the other bug! I'll dup my bug to #178 - do you want to pull my changes into your PR?

RickSBrown:
Ok I'll incorporate your changes into my PR.

PR #201, opened 2015-07-26
Don't need to take container scroll into account if container is body

### LINKED ISSUE #243: False positive for outsideScrollArea when body has margin
See http://robdodson.github.io/udacity-a11y/lesson2-focus/02-dom-order/ - the body element has a 60px margin at the bottom to accommodate the newsletter signup UI, which causes the focusable elements in that area to be "outside of scroll area".

alice:
Fixes #243.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -14,6 +14,7 @@
 * Revert #150 which was causing the extension not to work.
 * AX_HTML_02 (duplicate IDs) now only audits elements that are referenced by an IDREF (#141);
 * Fix #171 by being smarter about finding the composed parent node.
+* Tweak in canScrollTo to handle the (common) case where the container is `document.body` (#243).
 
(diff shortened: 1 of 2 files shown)


RickSBrown:
@alice Further to our slack discussion, I did some experiments with this and can confirm that it's highly infuriating to try to unit test. One of the biggest problems that took me ages to realise is that this bug only occurs in Chrome and Safari - I'd say webkit but it doesn't happen in PhantomJS. It also does not occur in Firefox or IE11.

In Firefox I'm pretty sure the bug doesn't occur because `axs.utils.canScrollTo` always returns `true` - even on an absolutely positioned element -9999px offscreen.

For what it's worth here's what I got up to today https://github.com/RickSBrown/accessibility-developer-tools/tree/testBodyScroll It's probably useless but maybe you can get something out of it.

PR #206, opened 2015-07-27
Ignore noscript branches in low contrast tests

### LINKED ISSUE #203: HTML noscript element should not be considered in the low contrast tests
`script` elements are already ignored, `noscript` should be too since there is no way to know how the content would be presented if JS was turned off.

This test should pass but currently fails:
```javascript
test("Noscript = no relevant elements", function() {
  var noscript = document.createElement("noscript");
  var para = noscript.appendChild(document.createElement("p"));
  para.textContent = "Please enable JS to use this app";
  para.style.backgroundColor = 'white';
  para.style.color = 'white';  // this line is optional
  fixture.appendChild(noscript);
  deepEqual(
    axs.AuditRules.getRule('lowContrastElements').run({ scope: fixture }),
    { result: axs.constants.AuditResult.NA }
  );
});
```

RickSBrown:
Closes #203

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -14,6 +14,7 @@
 * Add a special case to handle color `"transparent"` to fix (#180).
 * Fix `matchSelector` not working properly in browser environments without vendor prefixes (#189).
 * Fix false positives on elements with no role for Unsupported ARIA Attribute rule (#178 and #199).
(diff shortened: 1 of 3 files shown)


RickSBrown:
@alice PTAL

Issue #105, opened 2014-08-26
Run property auto-generation scripts as part of a grunt target


alice:
https://github.com/GoogleChrome/accessibility-developer-tools/blob/master/scripts/parse_aria_schemas.py and https://github.com/RickSBrown/accessibility-developer-tools/blob/master/scripts/aria_rdf_to_constants.xsl - anything else?

RickSBrown:
This task should also include investigating the feasibility of, and implementing:

1. 'Dumbing down' the scripts to the point that we delete them and instead do '1 for 1' transformation from XML to JS/JSON (using something like http://www.bramstein.com/projects/xsltjson/).

2. Generating a separate closure "module" for each input file and then relying on closure compiler / grunt to include them in the final output script as part of the standard concatenation routine.

PR #207, opened 2015-07-27
Fix ARIA tablist and ARIA tab scope

### LINKED ISSUE #204: ARIA tablist fails ariaRoleNotScoped audit
`tablist` is not required to be contained within any other ARIA role but this test fails:

```javascript
test('Scope role present - tablist', function() {
    var rule = axs.AuditRules.getRule('ariaRoleNotScoped');
    var container = fixture.appendChild(document.createElement('ul'));
    container.setAttribute('role', 'tablist');
    for (var i = 0; i < 4; i++) {
        var item = container.appendChild(document.createElement('li'));
        item.setAttribute('role', 'tab');
    }

    var actual = rule.run({ scope: fixture });
    equal(actual.result, axs.constants.AuditResult.PASS);
    deepEqual(actual.elements, []);
});
```

RickSBrown:
Closes #204 
The required context role for `tab` ended up on `tablist` instead.

This motivates me to get onto #105

### DIFF
--- a/test/audits/aria-role-not-scoped-test.js
+++ b/test/audits/aria-role-not-scoped-test.js
@@ -98,9 +98,10 @@ test('Scope role present - tablist', function() {
 test('Scope role missing - tab', function() {
     var rule = axs.AuditRules.getRule('ariaRoleNotScoped');
     var fixture = document.getElementById('qunit-fixture');
+    var container = fixture.appendChild(document.createElement('ul'));
     var expected = [];
     for (var i = 0; i < 4; i++) {
-        var item = fixture.appendChild(document.createElement('span'));
+        var item = container.appendChild(document.createElement('li'));
         item.setAttribute('role', 'tab');
         expected.push(item);
     }
--- a/test/audits/aria-role-not-scoped-test.js
+++ b/test/audits/aria-role-not-scoped-test.js
@@ -94,3 +94,18 @@ test('Scope role present - tablist', function() {
     equal(actual.result, axs.constants.AuditResult.PASS);
     deepEqual(actual.elements, []);
 });
+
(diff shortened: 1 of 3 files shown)


RickSBrown:
@alice PTAL

test/audits/aria-role-not-scoped-test.js
```diff
@@ -79,3 +79,33 @@ test('Scope role missing', function() {
...
+    equal(actual.result, axs.constants.AuditResult.PASS);
+    deepEqual(actual.elements, []);
+});
+
+test('Scope role missing - tab', function() {
+    var rule = axs.AuditRules.getRule('ariaRoleNotScoped');
+    var fixture = document.getElementById('qunit-fixture');
+    var expected = [];
+    for (var i = 0; i < 4; i++) {
+        var item = fixture.appendChild(document.createElement('span'));
```
alice: Why does this test use `span`s where the previous uses a list? Would this test pass if a list was used?

alice:
:+1: modulo test weirdness

test/audits/aria-role-not-scoped-test.js
RickSBrown: It was just to save one line - this test doesn't need a container, the spans are added directly to the fixture. It didn't feel right to add a bunch of `li` elements without a `ul` or `ol`.

I can change it to list if you prefer - the test will still work.

alice: If you can be bothered, I think it would help avoid future confusion, but :+1: either way.

RickSBrown: Ok no problem - will make that change then merge it.

PR #208, opened 2015-07-28
Fix issue #202 - an HTML Button containing an img with alt text should pass ControlsWithoutLabel

### LINKED ISSUE #202: HTML button containing img with alt attribute should pass controlsWithoutLabel
I believe this should be fine from an a11y perspective:

`<button>Save</button>`

But it fails as the test below shows:

```
test('Button element with image with alt needs no label', function() {
    // Setup fixture

    var button = document.createElement('button');
    var img = button.appendChild(document.createElement('img'));
    img.setAttribute("alt", "Save");
    fixture.appendChild(button);

    var rule = axs.AuditRules.getRule('controlsWithoutLabel');
    equal(rule.run({ scope: fixture }).result,
          axs.constants.AuditResult.PASS);
});
```

RickSBrown:
Closes #202 
Pretty straightforward fix (I hope).
Spent much more time on unit tests than on the fix.
@alice PTAL

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -14,7 +14,7 @@
 * Add a special case to handle color `"transparent"` to fix (#180).
 * Fix `matchSelector` not working properly in browser environments without vendor prefixes (#189).
 * Fix false positives on elements with no role for Unsupported ARIA Attribute rule (#178 and #199).
-* HTML button containing img with alt attribute no passes controlsWithoutLabel (#202)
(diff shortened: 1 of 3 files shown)


alice:
A hearty :+1:
Sorry for the slow review!

RickSBrown:
Awesome! :smiley_cat:

PR #209, opened 2015-07-28
Tighten up axs.utils.isValidNumber

RickSBrown:
Handles the following edge cases:
* A signed positive number like `+1`
* A float missing its integer part like `.1`
* An integer starting with a leading zero like `01`

Added approximately one metric gajillion unit tests.

@alice PTAL No reported bug on this so I guess it's low priority.


test/js/utils-test.js
```diff
@@ -327,3 +327,126 @@ test("getRoles on element with abstract role.", function() {
...
+});
+
+test("with zero.", function() {
+    var actual = axs.utils.isValidNumber("0");
+    strictEqual(actual.value, 0, "Zero should be parsed");
+    ok(actual.valid, "Integer should be valid");
+});
+
+test("with leading zero integer.", function() {
+    var actual = axs.utils.isValidNumber("09");
```
alice: What about the octal case, e.g. 012 (= decimal 10)?

test/js/utils-test.js
```diff
@@ -327,3 +327,126 @@ test("getRoles on element with abstract role.", function() {
...
+
+test("with leading zero negative integer.", function() {
+    var actual = axs.utils.isValidNumber("-09");
+    strictEqual(actual.value, -9, "Integer should be parsed");
+    ok(actual.valid, "Integer should be valid");
+});
+
+test("with string starting with number.", function() {
+    var actual = axs.utils.isValidNumber("10 foo");
+    equal(actual.valid, false, "Number starting with string is not valid");
```
alice: Nit: starts with a number, not a non-number string

alice:
I'm not sure about the leading 0 cases because of javascript's weird octal thing. Otherwise this is good.

test/js/utils-test.js
RickSBrown: Yes worth considering - I had to think about that... I decided this is not a general purpose number parser but specific to our a11y needs, usually ARIA attributes in HTML (e.g. `aria-valuenow`) - I couldn't find one that allowed for octal and I doubt there are any screen readers that read "ten" when they encounter "012".

Even `parseInt` no longer treats leading zero as octal (tested in Chrome, Firefox and Safari).
`parseInt("012")` returns decimal 12.
There is a good writeup on it here: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/parseInt#Octal_interpretations_with_no_radix

So I think it's safe for us to assume leading zero is not octal - if someone has an octal number they should convert it to decimal before we see it in the a11y tools.

alice: Ok I'm convinced! Thanks for chasing that up.

PR #212, opened 2015-07-28
Disabled elements should be ignored by low contrast audit

### LINKED ISSUE #205: Disabled button fails the contrast ratio test - should it?
The default contrast ratio of disabled buttons (in Chrome and probably others) fails the low contrast tests.

I would argue that since the user does not / cannot interact with disabled elements they should be ignored for this (and perhaps other?) tests.

Another argument in favor of ignoring them is that if you style a disabled element to meet the contrast ratio audit then it will probably lose the `disabled` look and harm usability.

RickSBrown:
Closes #205 
@alice PTAL

### DIFF
--- a/src/js/AccessibilityUtils.js
+++ b/src/js/AccessibilityUtils.js
@@ -578,6 +578,35 @@ axs.utils.hasLabel = function(element) {
 
 /**
  * @param {Element} element An element to check.
+ * @return {boolean} True if the element is disabled.
+ */
+axs.utils.isElementDisabled = function(element) {
+    if(!element || element.nodeType !== Node.ELEMENT_NODE) {
+        return false;
+    }
+    var tagName = element.tagName.toUpperCase();
+    return (tagName in axs.constants.NATIVELY_DISABLEABLE && element.hasAttribute("disabled")) ||
+        element.getAttribute("aria-disabled") === "true";
+};
+
+/**
+ * @param {Element} element An element to check.
+ * @return {boolean} True if the element or one of its ancestors is disabled.
+ */
+axs.utils.isElementOrAncestorDisabled = function(element) {
+    if (axs.utils.isElementDisabled(element))
+        return true;
+    var parent = axs.utils.parentElement(element);
+    // fieldset disabled-ness does not flow to descendants of its first legend
+    if (parent && !axs.browserUtils.matchSelector(parent, "fieldset>legend:first-of-type")) {
+        return axs.utils.isElementOrAncestorDisabled(parent);
+    }
+    else
+        return false;
+};
+
+/**
+ * @param {Element} element An element to check.
  * @return {boolean} True if the element is hidden from accessibility.
  */
 axs.utils.isElementHidden = function(element) {
--- a/src/js/AccessibilityUtils.js
+++ b/src/js/AccessibilityUtils.js
@@ -605,12 +605,11 @@ axs.utils.isElementDisabled = function(element) {
             axs.browserUtils.matchSelector(element, 'fieldset>legend:first-of-type *')) {
         return false;
     }
-    var next = element;
-    do {
+    for (var next = element; next !== null; next = axs.utils.parentElement(next)) {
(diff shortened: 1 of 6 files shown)


src/js/AccessibilityUtils.js
```diff
@@ -577,6 +577,40 @@ axs.utils.hasLabel = function(element) {
...
 
 /**
+ * Determine if this element is disabled without considering the effect of a disabled ancestor.
+ * Disabled here means that the element should be considered disabled according to specification.
+ * This element may or may not be effectively disabled in practice as this is dependent on implementation.
+ * @param {Element} element An element to check.
+ * @return {boolean} True if the element is disabled.
+ */
+axs.utils.isElementDisabled = function(element) {
+    if(!element || element.nodeType !== Node.ELEMENT_NODE) {
```
alice: Given that this is being explicitly called with an `Element` object, we probably don't need to check `nodeType`.

src/js/AccessibilityUtils.js
```diff
@@ -577,6 +577,40 @@ axs.utils.hasLabel = function(element) {
...
+ * @return {boolean} True if the element or one of its ancestors is disabled.
+ */
+axs.utils.isElementOrAncestorDisabled = function(element) {
+    if (axs.utils.isElementDisabled(element))
+        return true;
+    var parent = axs.utils.parentElement(element);
+    // fieldset disabled-ness does not flow to descendants of its first legend
+    if (parent && !axs.browserUtils.matchSelector(parent, "fieldset>legend:first-of-type")) {
+        return axs.utils.isElementOrAncestorDisabled(parent);
+    }
```
alice: This should be in our style guide once we have one, but please keep {} consistent between if and else, to avoid confusion.

alice: (actually, the `else` is unnecessary here anyway.)

src/js/AccessibilityUtils.js
```diff
@@ -577,6 +577,40 @@ axs.utils.hasLabel = function(element) {
...
+ * @param {Element} element An element to check.
+ * @return {boolean} True if the element or one of its ancestors is disabled.
+ */
+axs.utils.isElementOrAncestorDisabled = function(element) {
+    if (axs.utils.isElementDisabled(element))
+        return true;
+    var parent = axs.utils.parentElement(element);
+    // fieldset disabled-ness does not flow to descendants of its first legend
+    if (parent && !axs.browserUtils.matchSelector(parent, "fieldset>legend:first-of-type")) {
+        return axs.utils.isElementOrAncestorDisabled(parent);
```
alice: I think this may need a little more subtlety - e.g. a `<span role=button>` inside a `<fieldset disabled>` _won't_ be disabled, right? You'd need to add `aria-disabled` somewhere in the ancestor chain.

RickSBrown: Yeah good point. I'll need to add some unit tests for this... Will update shortly.

src/js/AccessibilityUtils.js
```diff
@@ -577,6 +577,44 @@ axs.utils.hasLabel = function(element) {
...
+axs.utils.isElementDisabled = function(element) {
+    if (axs.browserUtils.matchSelector(element, '[aria-disabled=true], [aria-disabled=true] *')) {
+        return true;
+    }
+    if(!axs.utils.isNativelyDisableable(element) ||
+            axs.browserUtils.matchSelector(element, 'fieldset>legend:first-of-type *')) {
+        return false;
+    }
+    var next = element;
+    do {
```
alice: This is fine, but why not
```for (var next = element; next !== null; next = axs.utils.parentElement(next) { ... }``` ?

RickSBrown: Love it!
I have no excuse for not coming up with that myself :( 

It's not a pattern I have got into the habit of using but it works so elegantly here I'm going to have to review all the JS I have written at work and look for places where I should have done this.

alice:
:+1:

PR #213, opened 2015-07-29
Link with unclear purpose audit handles img with alt and aria-label

RickSBrown:
This PR allows links to be labelled by their textcontent (as before) and additionally:

* aria-label
* aria-labelledby
* img with alt text

There is no way for us to truly validate the the link is *visually* meaningful since it would be really hard to analyse what the image looks like - this really only confirms that the link is meaningful to AT.

@alice PTAL

### DIFF
--- a/test/audits/link-with-unclear-purpose.js
+++ b/test/audits/link-with-unclear-purpose.js
@@ -27,6 +27,63 @@ test('a link with meaningful text is good', function() {
         { elements: [], result: axs.constants.AuditResult.PASS });
 });
 
+test('a link with an img with meaningful alt text is good', function() {
+    var fixture = document.getElementById('qunit-fixture');
+    var a = document.createElement('a');
+    a.href = '#main';
+    var img = a.appendChild(document.createElement('img'));
+    img.setAttribute('alt', 'Learn more about trout fishing');
+
+    fixture.appendChild(a);
+    var rule = axs.AuditRules.getRule('linkWithUnclearPurpose');
+    deepEqual(
+        rule.run({scope: fixture}),
+        { elements: [], result: axs.constants.AuditResult.PASS });
+});
+
+test('a link with an img with meaningless alt text is bad', function() {
+    var fixture = document.getElementById('qunit-fixture');
+    var a = document.createElement('a');
+    a.href = '#main';
+    var img = a.appendChild(document.createElement('img'));
+    img.setAttribute('alt', 'Click here!');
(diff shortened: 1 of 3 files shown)


test/audits/link-with-unclear-purpose.js
```diff
@@ -27,6 +27,67 @@ test('a link with meaningful text is good', function() {
...
+    fixture.appendChild(a);
+    var rule = axs.AuditRules.getRule('linkWithUnclearPurpose');
+    deepEqual(rule.run({scope: fixture}),
+        { elements: [a], result: axs.constants.AuditResult.FAIL });
+});
+
+test('a link with meaningful aria-label is good', function() {
+    var fixture = document.getElementById('qunit-fixture');
+    // Style our link to be visually meaningful with no descendent nodes at all.
+    fixture.innerHTML = '<style>a.trout::after{content: "Learn more about trout fishing"}</style>';
```
alice: Oh wow - I wonder if the a11y APIs pick up the ::after content - if they do, we should probably check for that as well (though that can definitely be a separate change).

alice: Ok, it looks like they do. I'll file a bug - don't worry about it in this change, but it might be worth making a note here that this test will need to be updated.

alice:
:+1: modulo the `::after` issue. Love the tests.

RickSBrown:
@alice I'll add that comment.
Glad you like the tests :)

I was just wondering if I should first add this test before I merge the PR?

It's the same as the content test except with a background image for visual meaning.

```javascript
test('a link with bg image and meaningful aria-label is good', function() {
    // Style our link to be visually meaningful with no descendent nodes at all.
    fixture.innerHTML = '<style>a.trout{background-image: url("/trout.png"); display:block; height:1em}</style>';
    var a = document.createElement('a');
    a.href = '#main';
    a.className = 'trout';
    a.setAttribute('aria-label', 'Learn more about trout fishing');
    fixture.appendChild(a);
    var rule = axs.AuditRules.getRule('linkWithUnclearPurpose');
    deepEqual(
        rule.run({scope: fixture}),
        { elements: [], result: axs.constants.AuditResult.PASS });
});
```

While it passes (and to me it makes sense that it does because the link is accessible visually and to AT) I wanted to run it by you first to see if this is expected behavior.

alice:
Yep, that looks fine to me.
:shipit:

Issue #217, opened 2015-07-30
findTextAlternatives not ignoring hidden element


ricksbrown:
For the anchor in the example below `findTextAlternatives` is returning “X Close this window" instead of the expected "Close this window".

```
<a href="#">
    <span aria-hidden="true”>X</span>
    <span>
        Close this window
    </span>
</a>
```

Reported here by @jdan https://web-a11y.slack.com/archives/a11y-devtools/p1438287595000326

PR #219, opened 2015-07-30
Fixes issue #217 findTextAlternatives not ignoring hidden elements

ricksbrown:
This is kind of scary because `findTextAlternatives` is a big and complex routine and I'm worried that the proposed change might have implications I have not considered.

That said, it passes all existing and new unit tests which gives me enough confidence to say @alice PTAL.

### DIFF
--- a/src/js/Properties.js
+++ b/src/js/Properties.js
@@ -386,7 +386,7 @@ axs.properties.findTextAlternatives = function(node, textAlternatives, opt_recur
         if (role && (!role.namefrom || role.namefrom.indexOf('contents') < 0))
             canGetNameFromContents = false;
     }
-    var textFromContent = axs.properties.getTextFromDescendantContent(element);
+    var textFromContent = axs.properties.getTextFromDescendantContent(element, opt_force);
     if (textFromContent && canGetNameFromContents) {
         var textFromContentValue = {};
         textFromContentValue.type = 'text';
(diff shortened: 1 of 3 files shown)


src/js/Properties.js
```diff
@@ -243,7 +243,7 @@ axs.properties.findTextAlternatives = function(node, textAlternatives, opt_recur
 
     // 1. Skip hidden elements unless the author specifies to use them via an aria-labelledby or
     // aria-describedby being used in the current computation.
-    if (!recursive && !opt_force && axs.utils.isElementOrAncestorHidden(element))
+    if (!opt_force && axs.utils.isElementOrAncestorHidden(element))
```
alice: How does this work? opt_force isn't passed in from the getTextFromAriaLabelledby method: https://github.com/GoogleChrome/accessibility-developer-tools/blob/48a69f03f644f9f83b3370e4a72d07b6d704bb1e/src/js/Properties.js#L475

alice: Ping? :)

ricksbrown: Sorry got sidetracked on another project :blush: 

The case I have added a test for is this:
```
<a href="#" aria-labelledby="foobar">
    <span id="foobar" aria-hidden="true">X</span>
    <span>Close this window</span>
</a>
```
To answer your question, it works by finding its text alternative not from line 475 but from here:
https://github.com/GoogleChrome/accessibility-developer-tools/blob/48a69f03f644f9f83b3370e4a72d07b6d704bb1e/src/js/Properties.js#L477

Actually I think line 477 might be a bug anyway and should read like this instead:
`labelledbyText.push(labelledby.text);` Which would then, indeed, result in this code not working. A test for this would be:
```
<span id="foobar" aria-label="Learn more about trout fishing"></span>
<a href="#" aria-labelledby="foobar"></a>
```

I think `opt_force` should always be passed in from the getTextFromAriaLabelledby method. A case to prove this would be:
```
<span id="foobar" aria-label="Learn more about trout fishing" aria-hidden="true"></span>
<a href="#" aria-labelledby="foobar"></a>
```
Interestingly:
ChromeVox does find the text alternative
Voiceover doesn't find the text alternative
chrome://accessibility/ doesn't find the text alternative

alice:
:+1:

PR #220, opened 2015-08-01
Automate style checking

### LINKED ISSUE #216: Automate style checking
I recommend running a linting tool as part of the grunt build. This would help reduce the overhead of reviewing PRs.

In my experience the best by far is ESLint: http://eslint.org/ 
Closure Linter would be a good fit too but is not as configurable.

ricksbrown:
To start with I have done minimal cleanup.

Eventually we should aim to address all the warnings and possibly raise the rule severity so the build fails if any of these issues are reintroduced.

We could optionally expand the checks to include test code.

@alice PTAL

.eslintrc configures the rules. `0` turns a rule off, `1` raises a warning, `2` raises an error.
Rules listed at http://eslint.org/docs/rules/ which links to detailed documentation.

Closes #216.


alice:
:+1: I'm going to pull this right now, amazing

Issue #141, opened 2015-03-19
IDREFs that refer to an ambiguous ID should be an error


dmazzoni-google:
Some accessibility audits give a warning if you have two elements on the page with the same id even if it's never used, and I think that's silly. It just wastes people's time in cases where it's not actually being used.

However, I think that code like this should be an error, since there's a label with an IDREF and there are multiple controls with that same id.

```
<label for="foo">First</label>
<input id="foo" value="First box">

<label for="foo">Second</label>
<input id="foo" value="Second box">
```

stevefaulkner:
@dmazzoni-google your code examples are not displaying, but agree that flagging mulitple id's as an issue is generally not an accessibility issue, though it is a HTML conformance error. 

The twist is that depending on the state of a document, an id value may or not be referenced in a relation so it is problematic to know when duplicate id's are problematic or not.

RickSBrown:
Alice and I discussed this a while back (see #102) and ended up writing an [audit](https://github.com/GoogleChrome/accessibility-developer-tools/blob/master/src/audits/DuplicateId.js) to cover this.

In terms of whether to only audit "used" ids or determine which ones are relevant to a11y I think this comment from the audit code explains our reasoning on it:

```
/*
 * Checks for duplicate IDs within the context of this element.
 * This is not a pure a11y check however IDREF attributes in ARIA and HTML (label 'for', td 'headers)
 *    depend on IDs being correctly implemented.
 * While we could limit this audit to IDs which are actually referred to via any IDREF attribute
 *    that would be incomplete, because naturally many of these attributes change over time e.g.
 *    'aria-activedescendant' is likely to change over time and therefore safer to vet them all.
 */
```

It's basically exactly what @stevefaulkner said.

FYI I fixed up the original comment from @dmazzoni-google so that the code example would display.

alice:
Since this test is so noisy I'm now convinced by @dmazzoni-google's reasoning.

IDREFs in the same scope really should be unique, otherwise there is no point having them, but it is indeed the case that automatically generated code often generates non-unique IDREFs and it is harmless unless something attempts to refer to that ID. Since there is a finite number of situations in which IDREFs may be used, perhaps we should follow @dmazzoni-google's original suggestion of finding all the cases where an IDREF is specified and make sure that each one refers to exactly one element (unless the element is aria-hidden or not rendered).

PR #221, opened 2015-08-02
Closes issue #141 - only audit IDs in IDREFs

ricksbrown:
Implements the decision in #141 
@alice PTAL

### DIFF
--- a/src/audits/AriaRoleNotScoped.js
+++ b/src/audits/AriaRoleNotScoped.js
@@ -54,7 +54,7 @@ axs.AuditRules.addRule({
         }
         // If we made it this far then no DOM ancestor has a required scope role.
         // Now we need to check if anything aria-owns this element.
-        var owners = axs.utils.getIdReferrers('aria-owns', element);  // there can only be ONE explicit owner but that's a different test
+        var owners = axs.utils.getAriaIdReferrers(element, 'aria-owns');  // there can only be ONE explicit owner but that's a different test
         if (owners) {
             for (var i = 0; i < owners.length; i++) {
                 var ownerRole = axs.utils.getRoles(owners[i], true);
--- a/src/audits/DuplicateId.js
+++ b/src/audits/DuplicateId.js
@@ -22,7 +22,7 @@ goog.require('axs.utils');
  */
 axs.AuditRules.addRule({
     name: 'duplicateId',
-    heading: 'An element\'s ID must be unique in the DOM',
+    heading: 'Any ID referred to via an IDREF must be unique in the DOM',
     url: 'https://github.com/GoogleChrome/accessibility-developer-tools/wiki/Audit-Rules#ax_html_02',
     severity: axs.constants.Severity.SEVERE,
     relevantElementMatcher: function(element) {
--- a/src/audits/DuplicateId.js
+++ b/src/audits/DuplicateId.js
@@ -28,7 +28,9 @@ axs.AuditRules.addRule({
     relevantElementMatcher: function(element) {
         if (element.hasAttribute('id')) {
             var referrers = axs.utils.getIdReferrers(element);
-            return referrers.length > 0;
+            return referrers.some(function(referrer) {
+                return !axs.utils.isElementOrAncestorHidden(referrer);
+            });
         }
         return false;
     },
@@ -43,16 +45,7 @@ axs.AuditRules.addRule({
         var id = element.id;
         var selector = '[id=\'' + id.replace(/'/g, '\\\'') + '\']';
         var elementsWithId = element.ownerDocument.querySelectorAll(selector);
-        if (elementsWithId.length > 1) {
-            var visible = [];
-            for (var i = 0; i < elementsWithId.length; i++) {
-                if (!axs.utils.isElementOrAncestorHidden(elementsWithId[i])) {
-                    visible.push(elementsWithId[i]);
-                }
-            }
-            return visible.length > 1;
-        }
-        return false;
+        return (elementsWithId.length > 1);
     },
     code: 'AX_HTML_02'
 });
--- a/src/audits/DuplicateId.js
+++ b/src/audits/DuplicateId.js
@@ -15,6 +15,7 @@
 goog.require('axs.AuditRules');
 goog.require('axs.browserUtils');
 goog.require('axs.constants.Severity');
+goog.require('axs.utils');
 
 /**
  * This audit checks for duplicate IDs in the DOM.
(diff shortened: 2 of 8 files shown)


src/audits/AriaRoleNotScoped.js
```diff
@@ -54,7 +54,7 @@ axs.AuditRules.addRule({
         }
         // If we made it this far then no DOM ancestor has a required scope role.
         // Now we need to check if anything aria-owns this element.
-        var owners = axs.utils.getIdReferrers('aria-owns', element);  // there can only be ONE explicit owner but that's a different test
+        var owners = axs.utils.getAriaIdReferrers(element, 'aria-owns');  // there can only be ONE explicit owner but that's a different test
```
alice: Why this change?

src/audits/DuplicateId.js
```diff
@@ -25,24 +26,33 @@ axs.AuditRules.addRule({
...
-            return false;  // Is an empty ID even an ID? I think not.
-        }
-        var selector = "[id='" + id.replace(/'/g, "\\'") + "']";
+        var selector = '[id=\'' + id.replace(/'/g, '\\\'') + '\']';
         var elementsWithId = element.ownerDocument.querySelectorAll(selector);
-        return (elementsWithId.length > 1);
+        if (elementsWithId.length > 1) {
+            var visible = [];
+            for (var i = 0; i < elementsWithId.length; i++) {
+                if (!axs.utils.isElementOrAncestorHidden(elementsWithId[i])) {
```
alice: I'm not sure about this part. I can kind of see the intent, but I think in practice the browser literally just picks the "first" element with the given ID whether it's visible or not. I just checked and this is certainly what Chrome does: https://code.google.com/p/chromium/codesearch#chromium/src/third_party/WebKit/Source/modules/accessibility/AXNodeObject.cpp&sq=package:chromium&type=cs&l=568&rcl=1438946804

src/audits/AriaRoleNotScoped.js
RickSBrown: I changed the API:
1. `axs.utils.getIdReferrers` now returns both HTML and ARIA IDREFs
2. The old "just return ARIA IDREFs" function is renamed to `getAriaIdReferrers`
3. I had to reverse the arguments because arg `attributeName` is now optional `opt_attributeName` - I'm operating on the practice of always putting optional arguments last. We could leave it the other way and pass it `null` if you prefer.

src/audits/DuplicateId.js
RickSBrown: @alice I was trying to implement your decision here of:
> - https://github.com/GoogleChrome/accessibility-developer-tools/issues/141#issuecomment-126779515 finding all the cases where an IDREF is specified and make sure that each one refers to exactly one element (unless the element is aria-hidden or not rendered).

The line you are commenting on is my attempt to do the "aria-hidden or not rendered" part but I probably misunderstood. Wait, I get it now, it makes sense - you meant if the element with the IDREFs is hidden, not the elements with the ID.

src/audits/AriaRoleNotScoped.js
alice: Ah, got it, thanks!

src/js/AccessibilityUtils.js
```diff
@@ -1041,30 +1041,99 @@ axs.utils.getQuerySelectorText = function(obj) {
...
+ * @return {NodeList} The elements that refer to this element.
+ */
+axs.utils.getHtmlIdReferrers = function(element) {
+    if (!element)
+        return null;
+    var id = element.id;
+    if (!id)
+        return null;
+    id = id.replace(/'/g, "\\'");  // make it safe to use in a selector
+    var selectorTemplates = ['[contextmenu=\'{id}\']', '[itemref~=\'{id}\']', 'button[form=\'{id}\']',
```
alice: Sorry, got distracted halfway through reviewing this! Could you make these one per line? I realise it'll be a lot of lines (20 or so?), but it's pretty hard to read like this (plus it'll make it easier to add any new ones if we realise something's missing).

src/audits/DuplicateId.js
alice: Yep, exactly - if an element <A> refers to another element <B>, but you can't see element <A>, it doesn't really matter.

src/js/AccessibilityUtils.js
RickSBrown: Ok, will do.

src/js/AccessibilityUtils.js
```diff
@@ -1041,30 +1041,113 @@ axs.utils.getQuerySelectorText = function(obj) {
...
+ */
+axs.utils.getHtmlIdReferrers = function(element) {
+    if (!element)
+        return null;
+    var id = element.id;
+    if (!id)
+        return null;
+    id = id.replace(/'/g, "\\'");  // make it safe to use in a selector
+    var selectorTemplates = [
+        '[contextmenu=\'{id}\']',
```
alice: Would be great to comment the source of this list so we can periodically check that it's up to date.

src/js/AccessibilityUtils.js
```diff
@@ -1041,30 +1041,113 @@ axs.utils.getQuerySelectorText = function(obj) {
...
+axs.utils.getIdReferrers = function(element) {
+    var result = [];
+    var addReferrers = function() {
+        for (var i = 0; i < referrers.length; i++) {
+            result.push(referrers[i]);
+        }
+    };
+    var referrers = axs.utils.getHtmlIdReferrers(element);
+    if (referrers) {
+        addReferrers();
```
alice: s/addReferrers()/Array.prototype.push.apply(result, referrers)/ ?

src/audits/DuplicateId.js
```diff
@@ -25,22 +26,24 @@ axs.AuditRules.addRule({
     url: 'https://github.com/GoogleChrome/accessibility-developer-tools/wiki/Audit-Rules#ax_html_02',
     severity: axs.constants.Severity.SEVERE,
     relevantElementMatcher: function(element) {
-        return axs.browserUtils.matchSelector(element, '[id]');
+        if (element.hasAttribute('id')) {
+            var referrers = axs.utils.getIdReferrers(element);
```
alice: This makes sense, but makes the heading confusing. Perhaps "Any ID referred to via an IDREF must be unique in the DOM"?

test/audits/duplicate-id-test.js
```diff
@@ -24,21 +24,35 @@ test('No duplicate IDs', function() {
...
-    var element2 = fixture.appendChild(element.cloneNode());
+    fixture.appendChild(element.cloneNode());
 
     var actual = rule.run({ scope: fixture });
-    equal(actual.result, axs.constants.AuditResult.FAIL);
-    deepEqual(actual.elements, [element, element2]);
+    equal(actual.result, axs.constants.AuditResult.NA);
 });
 
 test('Single duplicate ID but it\'s in shadow DOM', function() {
```
alice: We should duplicate this test for the IDREF used case.

src/js/AccessibilityUtils.js
ricksbrown: @alice sorry I got sidetracked on other work.
Sorry there is no definitive list I can find for this.

src/js/AccessibilityUtils.js
ricksbrown: Love it alice, very smart.
Unfortunately PhantomJS (2.0.0) can't handle it:
`'[object NodeList]' is not a valid argument for 'Function.prototype.apply' (evaluating 'Array.prototype.push.apply')`

We'd have to convert the NodeList to an Array first which somewhat defeats the purpose of what you were going for which was a nice one-liner instead of needing a helper function.

I could change it anyway if you prefer that style, including the conversion to array.

test/audits/duplicate-id-test.js
ricksbrown: Yeah good call.

src/audits/DuplicateId.js
ricksbrown: yeah true that. i changed it.

ricksbrown:
@alice I've addressed your feedback and this should hopefully be good to go?

alice:
:+1: Go ahead and merge :smile:

Issue #225, opened 2015-08-13
"This element has an unsuported ARIA attribute" 


linuslagerhjelm:
I get this error when an input tag of type text has an "aria-required='true'". I've tracked this down and it occurs because role="textbox" is not explicitly set. However, according to the W3 recommendation, this role gets implicitly set and is "not recommended" to declare explicitly.

stevefaulkner:
correct, no need to set a role value for a native HTML element that supports the use of the native version of the state/property

linuslagerhjelm:
But when left out, the tool will generate an error. Wouldn't that be a bug in the tool? And would best practices be to just ignore the error?

stevefaulkner:
suggest tool needs to be fixed

RickSBrown:
@linuslagerhjelm Thanks for reporting this, it is definitely a bug in the tool
@stevefaulkner thanks for comments :smile:

craigfrancis:
Just to add, this also happens with other elements as well, e.g.

    <th scope="col" aria-sort="ascending">

---

And not sure if this is quite right, but also:

    <abbr title="Required" aria-label="Required">*</abbr>

Where the design of the form uses an asterisk to show which fields are required (the input fields also add the required attribute). Unfortunately the designer insists on not having a key to say what it means (on the basis that "everyone knows what it means, so it clutters the interface"), where aria-labelledby would be better. That said, this is still redundant with the title attribute.

RickSBrown:
This one: `<abbr title="Required" aria-label="Required">*</abbr>` is a duplicate of #178 and has already been fixed (but not deployed as part of a release yet) in PR #195 

The other one: `<th scope="col" aria-sort="ascending">` hmmm. I think what's happened is the draft W3C document we based these rules on has gone through a few more iterations and I need to revisit the entire mapping of HTML elements to implicit ARIA roles.

craigfrancis:
Thanks Rick, I really appreciate this tool.

And yes, aria-sort has been changed a few times.

I'm still not completely sure I'm using aria-sort correctly, as I'm not using JS to change the order, instead using a simple link in the `<th>` (the tables represent many thousands of records, using pagination, so JS won't be able to handle this without server support).

RickSBrown:
Craig - a great place to discuss specific a11y related issues with other developers and major league a11y experts is here: https://gitter.im/w3c/a11ySlackers

[...]

PR #226, opened 2015-08-13
Fixes issue #225

ricksbrown:
Fixed the reported bug and found another related one we had not encountered yet - input with no `type` is a text input.

Actually an input of any unknown type is a text input - we probably don't handle that, that should probably be a separate issue..? For example `<input type="foo"/>` I'm thinking that's an outlier and no point holding up this fix for it.

Also completely ignoring the fact that `aria-required` should not be used when an exactly equivalent native attribute is available - that should be a different audit anyway because SEVERE is a bit extreme in that case.


RickSBrown:
@alice PTAL

alice:
:+1:

Issue #228, opened 2015-08-20
Enforcing recommended attributes on widgets


ctaymor:
We'd love to add audits for the recommended relationships between components with aria-roles. For instance, we are looking at a tabs widget. The [w3c guidelines for the role tabpanel](http://www.w3.org/TR/wai-aria/roles#tabpanel) state it should be associated with a tab, with either the tab having an aria-controls or the tabpanel having an aria-labelledby.

This seems to be a slightly unique case, looking at the [w3c documentation on aria roles](http://www.w3.org/TR/wai-aria/roles), but here are a few similar-ish cases.
* [alertdialog should have an aria-describedby](http://www.w3.org/WAI/PF/aria/roles#alertdialog)
* [dialog should have a label through aria-label or aria-labelledby](http://www.w3.org/WAI/PF/aria/roles#dialog)
* [region should have an aria-labelledby](http://www.w3.org/WAI/PF/aria/roles#region)

Would you be interested in this as a tab-specific rule (since it's a slightly unusual case)?

The Pivotal UI Team
@atomanyih @ctaymor @d-reinhold @kennyw12 @matt-royal @nicw @stubbornella

PR #218, opened 2015-07-30
Add "dataTableHeadersMissing" audit rule.

atomanyih:
This rule tests for compliance with [Section 508 #1194.22g](http://www.section508.gov/content/learn/standards/quick-reference-guide#1194.22g)

We also added [an entry to the wiki](https://github.com/GoogleChrome/accessibility-developer-tools/wiki/Audit-Rules#ax_table_01)

@matt-royal

PR #229, opened 2015-08-21
New rule: tabpanel should be related to a tab

ctaymor:
per the recomendations from the WAI-ARIA guidelines on the role tabpanel
found here: http://www.w3.org/TR/wai-aria/roles#tabpanel

Addresses #228

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -75,6 +75,7 @@
 * Elements with ARIA roles must ensure required owned elements are present (`src/audits/RequiredOwnedAriaRoleMissing.js`)
 * Avoid positive integer values for tabIndex (`src/audits/TabIndexGreaterThanZero.js`)
 * This element has an unsupported ARIA attribute (`src/audits/UnsupportedAriaAttribute.js`)
+* A tabpanel should be related to a tab via aria-controls or aria-labelledby (`src/audits/UncontrolledTabpanel.js)
 
 ### Enhancements:
 * Add configurable blacklist phrases and stop words to LinkWithUnclearPurpose (#99)
--- a/Changelog.md
+++ b/Changelog.md
@@ -15,6 +15,7 @@
 ### New rules
 
 * A label element may not have labelable descendants other than its labeled control (`src/audits/MultipleLabelableElementsPerLabel.js`)
+* A tabpanel should be related to a tab via aria-controls or aria-labelledby (`src/audits/UncontrolledTabpanel.js)
 
 ### Enhancements
 
@@ -75,7 +76,6 @@
 * Elements with ARIA roles must ensure required owned elements are present (`src/audits/RequiredOwnedAriaRoleMissing.js`)
 * Avoid positive integer values for tabIndex (`src/audits/TabIndexGreaterThanZero.js`)
 * This element has an unsupported ARIA attribute (`src/audits/UnsupportedAriaAttribute.js`)
-* A tabpanel should be related to a tab via aria-controls or aria-labelledby (`src/audits/UncontrolledTabpanel.js)
 
 ### Enhancements:
 * Add configurable blacklist phrases and stop words to LinkWithUnclearPurpose (#99)
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,4 +1,5 @@
 ### New rules
+* A tabpanel should be related to a tab via aria-controls or aria-labelledby (`src/audits/UncontrolledTabpanel.js)
 
 ### Enhancements
 
@@ -15,7 +16,6 @@
 ### New rules
 
 * A label element may not have labelable descendants other than its labeled control (`src/audits/MultipleLabelableElementsPerLabel.js`)
-* A tabpanel should be related to a tab via aria-controls or aria-labelledby (`src/audits/UncontrolledTabpanel.js)
 
 ### Enhancements
 
--- a/src/audits/UncontrolledTabpanel.js
+++ b/src/audits/UncontrolledTabpanel.js
@@ -25,7 +25,7 @@ function labeledByATab(element) {
 }
 
 function controlledByATab(element) {
-    var controlledBy = document.querySelectorAll('[role="tab"], [controls="' + element.id + '"]')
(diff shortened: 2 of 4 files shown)


ctaymor:
And we can add this to the Wiki:

## AX_ARIA_13
##

**Good:**
```html
<!-- Good: each tabpanel is referenced by a aria-controls on a different tab -->
<div role="tablist">
  <div role="tab" aria-controls="tabpanel1">Tab 1</div>
  <div role="tab" aria-controls="tabpanel2">Tab 2</div>
</div>

<div role="tabpanel" id="tabpanel1"> Some content in the first tab panel</div>
<div role="tabpanel" id="tabpanel2"> Other stuff in the second tab panel</div>

<!-- Good: each tabpanel has an aria-labelledby for a different tab -->
<div role="tablist">
  <div role="tab" id="tab1">Tab 1</div>
  <div role="tab" id="tab2">Tab 2</div>
</div>

<div role="tabpanel" aria-labeledby="tab1"> Some content in the first tab panel</div>
<div role="tabpanel" aria-labeledby="tab2"> Other stuff in the second tab panel</div>
```

**Bad:**
```html
<!-- Bad: tabpanel is not labeled by and there is no tab which controls it -->
<div role="tablist">
  <div role="tab">Tab 1</div>
  <div role="tab">Tab 2</div>
</div>

<div role="tabpanel"> Some content in the first tab panel</div>
<div role="tabpanel"> Other stuff in the second tab panel</div>

<!-- Bad: tabpanel is labeled by something other than a tab and not controlled by a tab -->
<div role="tablist">
  <div role="tab">Tab 1</div>
</div>

<div role="tabpanel" aria-labeledby="notATab"> Some content in the first tab panel</div>

<h5 id="notATab">This describes the content of the tabpanel</h5>

```

An element with the `tabpanel` ARIA roles should be related to an element with the role `tab`. Either the tabpanel can have an `aria-labelledby` referencing the tab's id, or the tab can have an `aria-controls` referencing the tabpanel's id.

You can find more information on the tabpanel role on the [WAI-ARIA site](http://www.w3.org/TR/wai-aria/roles#tabpanel).

dylanb:
This will generate mostly false positives for two reasons

1) it is only a SHOULD requirement, and
2) the tab panels can be added dynamically by JavaScript and only have to be there when the tab itself is selected

ctaymor:
The second concern shouldn't be a problem with this audit. It only checks that the tab panels have a tab which controls them, not whether the tab has a tab panel. So if the tab panel hasn't been added yet, it won't fail the test.

In regards to the first concern, I gave it a warning severity instead of severe for that reason. However, my understanding is that although this is only a SHOULD requirement, it is of great importance for the usability of tabs, and thus worth having an audit for.

test/audits/uncontrolled-tabpanel-test.js
```diff
@@ -0,0 +1,132 @@
...
+  var tab = document.createElement('div');
+  tab.setAttribute('role', 'tab');
+  tab.setAttribute('aria-controls', 'tabpanelId');
+
+  tabList.appendChild(tab);
+  var tabPanel = document.createElement('div');
+  tabPanel.setAttribute('role', 'tabpanel');
+  tabPanel.setAttribute('id', 'tabpanelId');
+  fixture.appendChild(tabPanel);
+
```
dylanb: Should add a passing test where there is more than one tab, but only one tabpanel to ensure the false positives do not occur

ctaymor:
Good idea @dylanb. I added the test for that case (and caught a bug in the process).

src/audits/UncontrolledTabpanel.js
```diff
@@ -0,0 +1,46 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+goog.require('axs.AuditRules');
+goog.require('axs.browserUtils');
+goog.require('axs.constants.Severity');
+
+function labeledByATab(element) {
```
alice: Similarly to #218, could you scope these methods to this file and JSDoc them?

test/audits/uncontrolled-tabpanel-test.js
```diff
@@ -0,0 +1,162 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+module('UncontrolledTabpanel');
+
+test('No elements === no problems.', function() {
```
alice: I think this tab and the one below are redundant with the one on line 38.

alice: Hm, actually, I see the value in the one below, but this one is certainly redundant.

Changelog.md
```diff
@@ -59,6 +59,7 @@
 * Elements with ARIA roles must ensure required owned elements are present (`src/audits/RequiredOwnedAriaRoleMissing.js`)
 * Avoid positive integer values for tabIndex (`src/audits/TabIndexGreaterThanZero.js`)
 * This element has an unsupported ARIA attribute (`src/audits/UnsupportedAriaAttribute.js`)
+* A tabpanel should be related to a tab via aria-controls or aria-labelledby (`src/audits/UncontrolledTabpanel.js)
```
alice: As with #218, we'll need to move this Changelog entry up above the latest release line (sorry!)

alice:
:+1: once comments addressed and Changelog updated.

ctaymor:
Thanks Alice. I made those changes.

alice:
@ctaymor Did you update the Changelog too? You may need to pull the latest version.

ctaymor:
@alice Sorry, I did update the changelog, but not correctly because I hadn't pulled in the latest version. This should fix it with the latest changelog.

alice:
Thanks!

PR #227, opened 2015-08-13
Hidden links no longer fail meaningful link audit

kennyw12:
Links which were hidden were failing the LinkWithUnclearPurpose audit, because they were being audited, but findTextAlternatives wasn't returning their text.

The Pivotal UI Team
@atomanyih @ctaymor @d-reinhold @kennyw12 @matt-royal @nicw @stubbornella

PR #233, opened 2015-08-28
Changes how ignoreSelectors works to match the documented behavior

d-reinhold:
Based on our reading of the README, calling 'ignoreSelectors' on a
config object with a particular selector should ignore matching nodes
and all their descendents, instead of just the nodes found by the selector.

We believe this is the correct behavior, because the README says to use
ignoreSelectors in order to 'Ignore parts of the page for a particular audit rule',
not just the element matching the selector.

Signed-off-by: Caroline Taymor <ctaymor@pivotal.io>

### DIFF
--- a/src/js/AuditRule.js
+++ b/src/js/AuditRule.js
@@ -119,17 +119,34 @@ axs.AuditRule.prototype.addElement = function(elements, element) {
 };
 
 /**
+ * @param {Node} node
+ * @param {function(Element): boolean} matcher
+ * @param {Array<string>=} opt_ignoreSelectors The array of selectors to ignore, if any.
+ */
+axs.AuditRule.collectMatchingElements = function(node, matcher, collection, opt_ignoreSelectors) {
+    axs.AuditRule.collectMatchingElementsInternal(node, matcher, collection, null, opt_ignoreSelectors);
+};
+
+/**
  * Recursively collect elements which match |matcher| into |collection|,
  * starting at |node|.
  * @param {Node} node
  * @param {function(Element): boolean} matcher
  * @param {Array.<Element>} collection
- * @param {ShadowRoot=} opt_shadowRoot The nearest ShadowRoot ancestor, if any.
+ * @param {?ShadowRoot} opt_shadowRoot The nearest ShadowRoot ancestor, if any
+ * @param {Array<string>=} opt_ignoreSelectors The array of selectors to ignore, if any.
  */
-axs.AuditRule.collectMatchingElements = function(node, matcher, collection,
-                                                 opt_shadowRoot) {
-    if (node.nodeType == Node.ELEMENT_NODE)
+axs.AuditRule.collectMatchingElementsInternal = function(node, matcher, collection, opt_shadowRoot, opt_ignoreSelectors) {
+    if (node.nodeType == Node.ELEMENT_NODE) {
         var element = /** @type {Element} */ (node);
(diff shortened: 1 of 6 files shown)


dist/js/axs_testing.js
```diff
@@ -13,7 +13,7 @@
  * See the License for the specific language governing permissions and
```
alice: Please revert the change to the compiled file - we only update this for releases.

src/js/AuditRule.js
```diff
@@ -124,12 +124,20 @@ axs.AuditRule.prototype.addElement = function(elements, element) {
  * @param {Node} node
  * @param {function(Element): boolean} matcher
  * @param {Array.<Element>} collection
- * @param {ShadowRoot=} opt_shadowRoot The nearest ShadowRoot ancestor, if any.
+ * @param {Object=} options: optional parameters: shadowRoot = The nearest ShadowRoot ancestor, if any. ignoredSelectors = The array of selectors to ignore, if any.
```
alice: I'm finding conflating the optional opt_shadowRoot argument with options confusing. How about making this into two methods, along these lines:

/**
 * @param {Node} node
 * @param {function(Element): boolean} matcher
 * @param {Array<string>=} opt_ignoreSelectors
 */
axs.AuditRule.collectMatchingElements = function(node, matcher, opt_ignoreSelectors) {
    axs.AuditRule.collectMatchingElementsInternal(node, matcher, null, opt_ignoreSelectors);
};

/**
 * @param {Node} node
 * @param {function(Element): boolean}
 * @param {?ShadowRoot} shadowRoot The nearest ShadowRoot ancestor, if any
 * @param {Array<string=} opt_ignoreSelectors
 */
axs.AuditRule.collectMatchingElementsInternal = function(node, matcher, shadowRoot, opt_ignoreSelectors) {
...
}

alice:
Nice catch and solid tests! Just need to revert the generated file (we only update that when we do a release), and address the optional arg weirdness (feel free to propose an alternative to my solution if you don't like it).

kennyw12:
Sorry for the mess again. It should be good to go.

alice:
Looks like there are still some unrelated changes in here. It might be better to `git rebase -i` or even just `git reset upstream/master` (note: no `--hard`) and create one single change for this PR.

kennyw12:
Thank you. I did a `git reset upstream/master`. That may have affected the googlebot ¯_(ツ)_/¯

d-reinhold:
I signed it!

willnorris:
(trying to debug the CLA situation...)

@d-reinhold Are you at Pivotal also?  If so, what's your pivotal.io email?  

We ran into a similar case with #227 where @kennyw12 submitted a pull request with a commit authored by @ctaymor.  In that case, both of them were present in the git commit, one as author and one as committer.  I still need to update our CLA checker to handle that case, but at least the data is there.  In this pull request, we would have no way to look you up and verify that you are also at Pivotal, since you don't appear anywhere in the commit data.  It would certainly be easier if the commit author (@kennyw12 in this case) also opened the pull request.  But if that's not easily possible, we can try and come up with other solutions that may help prevent this CLA confusion in the future.

atomanyih:
We created a duplicate PR #241, but it looks like the CLA is being accepted on this PR now?

alice:
I fixed the merge conflicts and pulled this change in #244 - sorry for all the confusion, but @kennyw12's change is in there!

PR #237, opened 2015-09-04
Revert #150

### LINKED ISSUE #236: #150 causes issues in the extension
The iframe handling code in PR #150 causes breakages in the extension, so I'm going to revert it.

@davidhsv if you depend on this behaviour, would you be able to create a PR which puts this behaviour behind a configuration option which is off by default?

alice:
@ckundo PTAL

Closes #236


ckundo:
LGTM once CI is green. Closes #236

PR #241, opened 2015-09-16
changes ignoreSelectors to match documented behavior

kennyw12:
Based on our reading of the README, calling 'ignoreSelectors' on a
config object with a particular selector should ignore matching nodes
and all their descendents, instead of just the nodes found by the selector.

We believe this is the correct behavior, because the README says to use
ignoreSelectors in order to 'Ignore parts of the page for a particular audit rule',
not just the element matching the selector.


atomanyih:
We made this per discussion on #233, but since the CLA seems to have been accepted we're closing this one.

PR #246, opened 2015-10-15
Fix #245

### LINKED ISSUE #245: `linkWithUnclearPurpose` should only look at links, not `<a>` without `href`.
Ok:
```
<a id="this-is-an-anchor"></a>
```

Not ok:
```
<a id="this-is-a-link" href="/"></a>
```

alice:
@ricksbrown PTAL?

### DIFF
--- a/test/audits/link-with-unclear-purpose.js
+++ b/test/audits/link-with-unclear-purpose.js
@@ -132,3 +132,13 @@ test('a hidden link should not be run against the audit', function() {
       rule.run({scope: fixture}),
       { result: axs.constants.AuditResult.NA });
 });
+
+test('an anchor tag without href attribute is ignored', function() {
+    var fixture = document.getElementById('qunit-fixture');
+    fixture.appendChild(document.createElement('a'));
+
+    var rule = axs.AuditRules.getRule('linkWithUnclearPurpose');
+    deepEqual(
+      rule.run({scope: fixture}),
+      { result: axs.constants.AuditResult.NA });
+});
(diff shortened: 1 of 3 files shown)


test/audits/link-with-unclear-purpose.js
```diff
@@ -132,3 +132,13 @@ test('a hidden link should not be run against the audit', function() {
       rule.run({scope: fixture}),
       { result: axs.constants.AuditResult.NA });
 });
+
+test('an anchor tag without href attribute is ignored', function() {
+    var fixture = document.getElementById('qunit-fixture');
+    var a = fixture.appendChild(document.createElement('a'));
```
ricksbrown: totally nitpicking here but `var a` is never used.

test/audits/link-with-unclear-purpose.js
```diff
@@ -132,3 +132,13 @@ test('a hidden link should not be run against the audit', function() {
...
+
+    var rule = axs.AuditRules.getRule('linkWithUnclearPurpose');
+    deepEqual(
+      rule.run({scope: fixture}),
```
ricksbrown: style nit: space after `{` and before `}`

ricksbrown:
@alice couple of trifling issues.
This is a nice little patch, good stuff, LGTM.

test/audits/link-with-unclear-purpose.js
alice: Done.

test/audits/link-with-unclear-purpose.js
alice: Would fix this but it's consistent with the rest of the file, so I don't think it's worth it.

PR #248, opened 2015-11-16
Added some shiny badges

ricksbrown:
Given the huge number of downloads we should probably advertise this.

[Can see it working here](https://github.com/ricksbrown/accessibility-developer-tools/tree/3111851afd57db7d435b8b44735d5942fd51063b).


jdan:
Cool! LGTM

alice:
:+1: yay

loginx:
:+1: nice!

PR #250, opened 2015-11-18
added aria-multiselectable to tablist

ricksbrown:
[tablist](http://www.w3.org/TR/wai-aria/roles#tablist) should support [aria-multiselectable](http://www.w3.org/TR/wai-aria/states_and_properties#aria-multiselectable)


ricksbrown:
This is currently failing CI due to #301

ricksbrown:
fixes #289 
@alice PTAL

PR #259, opened 2015-11-26
adding layout table checking function

garcialo:


src/audits/TableHasAppropriateHeaders.js
```diff
@@ -73,6 +73,32 @@ goog.require('axs.constants.Severity');
...
+     *
+     * @returns {boolean} Table is a layout table
+     */
+    function tableIsAFilthyLayoutTable(element) {
+        if (element.getAttribute('role') == 'presentation') {
+            return true;
+        }
+
+        if (element.hasAttribute("summary")) {
+            return true;
```
alice: Summary probably indicates it _isn't_ a layout table?

src/audits/TableHasAppropriateHeaders.js
```diff
@@ -73,6 +73,32 @@ goog.require('axs.constants.Severity');
...
+        }
+			
+        // empty tables considered layout tables
+        if (!element.childElementCount > 0) {
```
alice: `!` is higher precedence than `>` so this amounts to `if ((!element.childElementCount) > 0)` which is probably not what you wanted (though it coincidentally ends up doing the right thing), but you probably want to just do `if (childElementCount == 0)` anyway.

src/audits/TableHasAppropriateHeaders.js
```diff
@@ -84,8 +110,8 @@ goog.require('axs.constants.Severity');
         },
         test: function (element) {
 
-            if (element.getAttribute('role') == 'presentation') {
-                return element.querySelectorAll('th').length != 0;
+            if (tableIsAFilthyLayoutTable(element)) {
```
alice: Did you intend to call this from `relevantElementMatcher()`?

src/audits/TableHasAppropriateHeaders.js
alice: Also, note that if the only element in the table is a `<thead>` or `<tbody>` then the table will have child elements, but no actual cells. Perhaps do a querySelector for `tr` or `td`?

README.md
```diff
@@ -200,7 +202,7 @@ which allows you to provide multiple configuration options at once.
 
 ## License
 
-Copyright 2013 Google Inc. All Rights Reserved.
+Copyright 2015 Google Inc. All Rights Reserved.
```
alice: No need to update the date, AIUI.

src/audits/TableHasAppropriateHeaders.js
```diff
@@ -73,6 +73,45 @@ goog.require('axs.constants.Severity');
...
+
+        if (element.getAttribute('role') == 'presentation') {
+            return true;
+        }
+
+        if (element.hasAttribute("summary")) {
+            return false;
+        }
+		
+		if (element.querySelectorAll('thead').length > 0) {
```
alice: The indenting is inconsistent - looks like you switched from spaces to tabs.

src/audits/TableHasAppropriateHeaders.js
```diff
@@ -73,6 +73,45 @@ goog.require('axs.constants.Severity');
...
+			return false;
+		}
+
+		if (element.querySelectorAll('th').length > 0) {
+			return false;
+		}
+
+		// check if any children of element have aria-label, aria-labelledby, aria-describedby, title
+
+		return false;
```
alice: There's a lot of `if (x) return false` and then `return false`?

src/audits/TableHasAppropriateHeaders.js
```diff
@@ -73,6 +73,45 @@ goog.require('axs.constants.Severity');
         }
         return false;
     }
+	
+    /**
+     * Checks whether a table is a layout table.
+     *
+     * @returns {boolean} Table is a layout table
+     */
+    function tableIsAFilthyLayoutTable(element) {
```
alice: The editorialising is funny, but I'm afraid I can't pull it :smile:

src/audits/TableHasAppropriateHeaders.js
```diff
@@ -73,6 +73,45 @@ goog.require('axs.constants.Severity');
...
+
+		if (element.querySelectorAll('tfoot').length > 0) {
+			return false;
+		}
+
+		if (element.querySelectorAll('th').length > 0) {
+			return false;
+		}
+
+		// check if any children of element have aria-label, aria-labelledby, aria-describedby, title
```
alice: Is there any code to go with this comment?

src/audits/TableHasAppropriateHeaders.js
```diff
@@ -80,20 +119,15 @@ goog.require('axs.constants.Severity');
...
 
                 return tableDoesNotHaveHeaderRow(rows) &&
                     tableDoesNotHaveHeaderColumn(rows) &&
                     tableDoesNotHaveGridLayout(rows);
-            }
         },
         code: 'AX_TABLE_01',
     });
-})();
+})();
```
alice: Needs a newline at the end of the file. Is NP++ stripping them?

src/audits/TableHasAppropriateHeaders.js
```diff
@@ -73,6 +73,34 @@ goog.require('axs.constants.Severity');
...
+     * Checks whether a table is a layout table.
+     *
+     * @returns {boolean} Table is a layout table
+     */
+    function isLayoutTable(element) {
+        if (element.childElementCount == 0) {
+            return true;
+        }
+
+        if (element.getAttribute('role') == 'presentation') {
```
alice: Consider early outing here along the lines of:
```js
if (!element.hasAttribute('role') || element.getAttribute('role') != 'presentation')
   return false;

// rest of code un-indented
```

src/audits/TableHasAppropriateHeaders.js
```diff
@@ -73,6 +73,34 @@ goog.require('axs.constants.Severity');
...
+            var tableChildren = element.querySelectorAll('*')
+
+            for (var i = 0; i < tableChildren.length; i++) {
+                if (tableChildren[i].tagName == 'TR' || tableChildren[i].tagName == 'TD') {
```
alice: Can you add a comment explaining this?

Also, I'd write it inverting the logic, i.e.:
```js
if (tableChildren[i].tagName != 'TR' && tableChildren[i].tagName != 'TD')
   return false;
```

test/audits/table-has-appropriate-headers-test.js
```diff
@@ -61,6 +61,19 @@ module('TableHasAppropriateHeaders');
         return table;
     }
 
+    test('Table with no child elements', function () {
```
alice: This is good, but we need a bunch more tests. At the very least, the original case we were trying to address, where we have a `<table>` with child elements but no `<tr>` descendants.

test/audits/table-has-appropriate-headers-test.js
```diff
@@ -61,6 +61,19 @@ module('TableHasAppropriateHeaders');
...
+        var rule = axs.AuditRules.getRule('tableHasAppropriateHeaders');
+        var fixture = document.getElementById('qunit-fixture');
+
+        var table = document.createElement('table');
+
+        fixture.appendChild(table);
+
+        var actual = rule.run({scope: fixture});
+        equal(actual.result, axs.constants.AuditResult.PASS);
```
alice: Shouldn't this be `axs.constants.AuditResult.NA`?

Issue #147, opened 2015-04-23
failure due to ReferenceError


jaimeiniesta:
When testing http://ludus.org.es/en the audit crashes and returns `null`. It looks like there's a JS error on that page and `axs_testing.js` can't stand it.

This happens both on the latest build from master and on the latest release from the stable branch.

I'm using phantomjs 2.0.0 and OSX Yosemite.

```js
$ phantomjs tools/runner/audit.js http://ludus.org.es/en

ReferenceError: Can't find variable: jQuery

  https://goteo.org/widget/project/ludus:62 in global code
ReferenceError: Can't find variable: jQuery

  https://goteo.org/widget/project/ludus:137 in global code
ReferenceError: Can't find variable: goteo

  https://goteo.org/widget/project/ludus:253 in global code
SyntaxError: Unexpected EOF

  ../../dist/js/axs_testing.js:1 in injectJs
  tools/runner/audit.js:33
  :285 in _onPageOpenFinished
ReferenceError: Can't find variable: axs

  phantomjs://webpage.evaluate():2
  phantomjs://webpage.evaluate():4
null
```

PR #260, opened 2015-11-27
fix: make phantom audit work with websites that expose an AMD loader

joscha:
In the case that the audited website has an exposed AMD loader, the auditor fails, because `window.axs` is undefined. This PR fixes this.

### DIFF
--- a/tools/runner/audit.js
+++ b/tools/runner/audit.js
@@ -33,6 +33,13 @@ if (system.args.length !== 2) {
         );
       phantom.exit();
     } else {
+      page.evaluate(function() {
+        // if target website has an AMD loader, we need to make sure
+        // that window.axs is still available
+        if (typeof define !== 'undefined' && define.amd) {
+            define.amd = false;
+        }
+      });
       page.injectJs('../../dist/js/axs_testing.js');
       var report = page.evaluate(function() {
         var results = axs.Audit.run();


alice:
@ckundo Would you like to take a look at this, given you wrote the PhantomJS runner?

@joscha Apologies for taking so long to look at this! I'd like to merge this, but I don't have enough understanding of either PhantomJS or AMD to grok what's going on here - any chance you could comment with a little explainer to point me in the right direction?

joscha:
@alice so the problem is, that `axs_testing.js` uses an AMD loader if one is available and only exposes the global `axs` if that is not the case (see https://github.com/GoogleChrome/accessibility-developer-tools/blob/master/dist/js/axs_testing.js#L2272). When a JS file gets injected with PhantomJS, it basically means that it runs within the page environment it was injected in. So if that page exposes an AMD environment (e.g. `define` is defined and `define.amd` is not falsy) the `axs_testing.js` does not register itself on the `window.axs` global, but anonymously registers itself via the `define` function. Subsequently, the code using it (e.g. the `axs.Audit` call within `page.evaluate`) can't access the global (because it was never registered) and fails. This fix sets `define.amd` to false, basically telling `axs_testing.js` that there is no viable AMD environment available, hence exposing the `window.axs` global.

joscha:
Fixes #147.

alice:
Great explanation, thanks!

Issue #263, opened 2015-12-03
Performance


alice:
The audit is slow to the point of unusable on large pages with many elements.

alice:
Idea: prune out parts of the page which are hidden from all users.
We might need to be judicious about this: the reason we didn't do this from the beginning is that hidden elements are usually just waiting around to be shown, and so not auditing them may fail to catch real accessibility issues which are lurking.
It might be worth auditing the rules to see which ones truly don't make sense to run on hidden elements.

ricksbrown:
@alice I agree with everything you say above.

I believe that the optimum performance would be attained if there was a "one touch" rule for each DOM element. Currently each rule queries the DOM independently, often evaluating the same nodes, for example `isElementOrAncestorHidden`.

Imagine an audit runner that essentially walks the DOM once, checks if the current element is hidden (disabled etc) and then passes that information to each rule along with the element.
In other words each rule would not know how to search the DOM for "relevant elements" but rather would take a single element, check it is "relevant" and then audit it.

Some rules will be difficult to achieve true "one touch" on because, for example, they may need to collect ID referrers or check what kind of parent or ancestor they are contained in, or what child nodes they have. Even this could probably be solved (some kind of crazy state machine perhaps) but I think that would probably not be necessary.

It would be a bit of work to get it working this way but I believe it would make a huge difference.

efrenamarille:
On 4 Dec 2015 15:31, Rick Brown <notifications@github.com> wrote:@alice I agree with everything you say above.

I believe that the optimum performance would be attained if there was a "one touch" rule for each DOM element. Currently each rule queries the DOM independently, often evaluating the same nodes, for example isElementOrAncestorHidden.

Imagine an audit runner that essentially walks the DOM once, checks if the current element is hidden (disabled etc) and then passes that information to each rule along with the element.
In other words each rule would not know how to search the DOM for "relevant elements" but rather would take a single element, check it is "relevant" and then audit it.

Some rules will be difficult to achieve true "one touch" on because, for example, they may need to collect ID referrers or check what kind of parent or ancestor they are contained in, or what child nodes they have. Even this could probably be solved (some kind of crazy state machine perhaps) but I think that would probably not be necessary.

It would be a bit of work to get it working this way but I believe it would make a huge difference.

efrenamarille:
[...]

PR #265, opened 2015-12-10
Performance

ricksbrown:
- Begins to address #263. There are definitely more optimizations to make but this is already a large PR.
- Addresses a TODO, namely expanding AX_ARIA_02 so that it covers HTML attributes as well as ARIA attributes.
- Fixed some style nits while I was there

Hopefully the performance improvement works on other pages (and hopefully you are happy with the changes).

Here are the results on my test page, using `console.time` (no audits disabled, includes DuplicateID):

Before: 7 seconds
After: 2 seconds (70% faster)

@alice PTAL and sorry it's a big one to review! There's about 40 hours work in it.

### DIFF
--- a/src/audits/DuplicateId.js
+++ b/src/audits/DuplicateId.js
@@ -25,14 +25,34 @@ axs.AuditRules.addRule({
     heading: 'Any ID referred to via an IDREF must be unique in the DOM',
     url: 'https://github.com/GoogleChrome/accessibility-developer-tools/wiki/Audit-Rules#ax_html_02',
     severity: axs.constants.Severity.SEVERE,
-    relevantElementMatcher: function(element) {
-        if (element.hasAttribute('id')) {
-            var referrers = axs.utils.getIdReferrers(element);
-            return referrers.some(function(referrer) {
-                return !axs.utils.isElementOrAncestorHidden(referrer);
+    opt_requires: {
+        idRefs: true
+    },
+    /**
+     * @this {axs.AuditRule}
+     */
+    relevantElementMatcher: function(element, flags) {
+        if (flags.idrefs.length && !flags.hidden) {
+            this.relatedElements.push({
+                element: element,
+                flags: flags
             });
         }
+        if (element.hasAttribute('id')) {
+            return true;
+        }
         return false;
+    },
+    /**
+     * @this {axs.AuditRule}
+     */
+    isRelevant: function(element, flags) {
+        var id = element.id;
+        var level = flags.level;
+        return this.relatedElements.some(function(related) {
+            var idrefs = related.flags.idrefs;
+            return related.flags.level === level && idrefs.indexOf(id) >= 0;
+        });
     },
     test: function(element) {
         /*
--- a/src/js/AccessibilityUtils.js
+++ b/src/js/AccessibilityUtils.js
@@ -549,10 +549,12 @@ axs.utils.isNativelyDisableable = function(element) {
  * This element may or may not be effectively disabled in practice as this is dependent on implementation.
  *
  * @param {Element} element An element to check.
+ * @param {boolean=} ignoreAncestors If true do not check for disabled ancestors.
  * @return {boolean} true if the element or one of its ancestors is disabled.
  */
-axs.utils.isElementDisabled = function(element) {
-    if (axs.browserUtils.matchSelector(element, '[aria-disabled=true], [aria-disabled=true] *')) {
+axs.utils.isElementDisabled = function(element, ignoreAncestors) {
+    var selector = ignoreAncestors ? '[aria-disabled=true]' : '[aria-disabled=true], [aria-disabled=true] *';
+    if (axs.browserUtils.matchSelector(element, selector)) {
         return true;
     }
     if (!axs.utils.isNativelyDisableable(element) ||
@@ -560,9 +562,12 @@ axs.utils.isElementDisabled = function(element) {
         return false;
     }
     for (var next = element; next !== null; next = axs.dom.parentElement(next)) {
-        if (axs.utils.isNativelyDisableable(next) && next.hasAttribute('disabled')) {
+        if (next.hasAttribute('disabled')) {
             return true;
         }
+        if (ignoreAncestors) {
+            return false;
+    }
     }
     return false;
 };
--- a/src/js/AccessibilityUtils.js
+++ b/src/js/AccessibilityUtils.js
@@ -567,7 +567,7 @@ axs.utils.isElementDisabled = function(element, ignoreAncestors) {
         }
         if (ignoreAncestors) {
             return false;
-    }
+        }
     }
     return false;
 };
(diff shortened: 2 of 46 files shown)


src/js/Audit.js
```diff
@@ -218,68 +218,122 @@ axs.Audit.getRulesCannotRun = function(opt_configuration) {
...
+            level: 0,
+            ignoring: {},
+            disabled: false,
+            hidden: false
+        };
+        var root = configuration.walkDom ? document.documentElement : configuration.scope;
+        // Because 'related elements' could occur anywhere in the DOM we need to start at document.documentElement
+        axs.dom.composedTreeSearch(root, null, { preorder: function(element, flags) {
+            if (!flags.inScope)
+                flags.inScope = element === configuration.scope;
```
alice: This will need to be undone when leaving the scope - you could add a `postorder` function to achieve this.

src/js/Audit.js
```diff
@@ -218,68 +218,122 @@ axs.Audit.getRulesCannotRun = function(opt_configuration) {
...
-    return results;
-};
-goog.exportSymbol('axs.Audit.run', axs.Audit.run);
-
+    /**
+     * Determine if this element should be ignored by the AuditRule.
+     * @param {Element} element An audit candidate.
+     * @returns {boolean} true if this element should be ignored by this AuditRule.
+     */
+    AuditOptions.prototype.ignore = function(element) {
```
alice: "ignore" here reads to me like the imperative; I would expect this to update the configuration to ignore the given element, based on the name.

What about something like `shouldIgnore` or `ignored` or `isIgnored`?

src/js/AuditRule.js
```diff
@@ -47,13 +48,26 @@ axs.AuditRule = function(spec) {
     /**
      * Is this element relevant to this audit?
      * @param {Element} element A potential audit candidate.
+     * @param {Object} flags
```
alice: As I read it, `flags` is actually optional. If this is the case, let's annotate it as such (`{Object=}`) and name accordingly to be consistent with other optional params.

src/js/AccessibilityUtils.js
```diff
@@ -549,10 +549,12 @@ axs.utils.isNativelyDisableable = function(element) {
  * This element may or may not be effectively disabled in practice as this is dependent on implementation.
  *
  * @param {Element} element An element to check.
+ * @param {boolean} self If true only consider the element itself.
```
alice: This param name is a little confusing, given `self` is often used to alias `this`. Perhaps something like `checkForInherited` (sense inverted)?

Also, this only limits the scope for the `aria-disabled` check, and doesn't prevent walking up the parent chain for native `disabled`. Is that deliberate? Perhaps early-out the for loop if this is true (or false if the sense is inverted, obviously)?

src/audits/DuplicateId.js
```diff
@@ -25,15 +25,36 @@ axs.AuditRules.addRule({
...
         }
+        if (element.hasAttribute('id')) {
+            return true;
+        }
         return false;
     },
+    /**
+     * @this {axs.AuditRule}
+     */
+    isRelevant: function(element, flags) {
```
alice: This is pretty hard to follow.

It seems like a bunch of infrastructure in this change is simply to make this work. I think it's definitely worthwhile, but it'll be much easier to review if you can split it into two or three changes - potentially one with just the formatting fixes (thanks for that btw), then one with the main performance optimisation (the single pass for collectMatchingElements, and related changes), and then one for this extra optimisation.

I know from experience how annoying this is to do, but git makes it tractable, and it'll make the review a lot easier. Feel free to ping me on hangouts if you need any help with the mechanics - not sure how much of this git surgery you've had to do in the past, but I've definitely done my fair share and happy to share what I've learnt!

src/js/AuditRule.js
```diff
@@ -47,13 +48,26 @@ axs.AuditRule = function(spec) {
...
      * @return {boolean} true if this element is relevant to this audit.
      */
     this.relevantElementMatcher_ = spec.relevantElementMatcher;
 
     /**
+     * If the relevantElementMatcher is not enough to determine whether or not this rule should run
+     *    then this function can be called (after DOM traversal is finished).
```
alice: Extra space before "then" isn't necessary - did the formatter add that?

alice:
On first glance, this looks really solid and I'm really excited about getting it in, but as you noted it's a lot to bite off in one review. How would you feel about breaking it into three changes as described in the comment above?

src/audits/DuplicateId.js
ricksbrown: Yep it's hard to follow, I should have commented it more and I'm glad you're looking at this while it's still in my head.

I can split it up, sure thing.
If you're in my neighbourhood feel free to come to my place and we could go over it together. It really does need explaining.

This particular line `isRelevant` is necessary to faithfully preserve the `DuplicateId` audit's behaviour:

- It needs to return `NA` if there are no relevant elements.
- Relevant elements are those which have one or more "id referrers".

The problem: ID referrers may be encountered **after** the element, so we have to wait until the DOM traversal is totally finished before we can know if there are *really* relevant elements.

Without `isRelevant` DuplicateId would return `PASS` instead of `NA` for every element with an ID.

src/js/Audit.js
ricksbrown: I may have stuffed this up but the way it's meant to work is that every level of the DOM tree actually has its own copy of the `flags` object which inherits from the parent flags object (using the prototype chain).

This happens in `composedTreeSearch` (sorry to taint your masterpiece) on this line:
`var flags = Object.create(parentFlags);`

So `flags.inScope` only affects the current level of DOM traversal and its entire subtree.
Well that's the theory anyway...

alice: Got it. Will take another look at this once you've managed to factor out the three separate changes.

src/js/AuditRule.js
ricksbrown: lol no - one Alice Boxhall [told me to do that](https://github.com/GoogleChrome/accessibility-developer-tools/pull/100#discussion_r16298003) :grinning:

src/js/Audit.js
ricksbrown: Agreed - will be in the bite-sized PR.

src/js/AuditRule.js
ricksbrown: It will always be there, I think it isn't optional? It will be easier to see when I give you the smaller PR.

src/js/AuditRule.js
alice: Ah - that's only for params and returns :)

ricksbrown: Ah right, got it, thanks!

ricksbrown:
@alice Hope this is OK, I've made it easier to pull parts out by splitting into three distinct commits.
Hopefully my git surgery was OK, thanks for the offer of help but I didn't want you to get frustrated at my git noobery :blush: 

We started with a PR that had three separate changes, 56 files changed, `+2,893 -2,006`

**Change 1, Stylistic Tweaks**
All gone (for now).

**Change 2, Audit Runner**
PART 1
The [core changes](https://github.com/ricksbrown/accessibility-developer-tools/commit/ef5cef634fc55423124ab0fb2001340a8e3c38a1) to the audit runner and a couple of tweaked AuditRules now account for: 
39 files changed  `+2,067 −1,157`. So that's literally thousands of changes you do not need to review, yay!

This does not deliver any significant performance boosts because the majority of the AuditRules are left untouched.

PART 2
[Updating the audit](https://github.com/ricksbrown/accessibility-developer-tools/commit/c282f927a81cdac0802a802dcd04526c5c6edbb0) rules adds:
12 changed files (eek!) but only `+31 -34` (phew!) 
These changes have a large impact impact on performance.

**Change 3, NonExistentRelatedElement**
I know you wanted a separate PR but it seemed very small: `+27 -38` so [I've thrown it in there](https://github.com/ricksbrown/accessibility-developer-tools/commit/272613e9f68398ec42fa8bf3ee7ad9c7dcfdd8aa).

Let me know if you still want me to remove this.

**Summary**
The grand total now stands at:
52 files changed `+2,118 −1,222`

Honestly the vast majority of changes are robotic cut and paste to the unit tests. Once you are over the complex core changes it should be smooth sailing.

I'd be happy to tune into hangouts or something and virtually sit with you while you go over it if you think it would help.

alice:
Hey - sorry to be this person, but what did you think of my suggestion about splitting this up even further?

ricksbrown:
As [discussed](https://web-a11y.slack.com/archives/a11y-devtools/p1450227033000083) git surgery pending - please hold off reviewing this until I post back here.

ricksbrown:
@alice git surgery complete - it is now split into seven commits which, if you review the changes in order, should hopefully make more sense.

The first commit is the biggest because it changes the unit tests.

I realised I did make some stylistic changes by encapsulating certain functionality into new "classes" like `AuditOptions` or new helpers like `axs.AuditRules.getActiveRules`. The functionality used to be there and I have essentially shuffled it around. If you hate it encapsulated that way I could probably undo it to some extent. I didn't give it a huge amount of thought, I was just trying to break things down, reduce cyclomatic complexity and scope so I could get my head around it and see what was used where.

alice:
Thank you SO MUCH for doing that git surgery. It made it so much easier to review and to see each idea in isolation. I really love this work - obviously I had a bunch of comments but no show stoppers; I'm really excited to get this in.

addyosmani:
@ricksbrown This work looks 🏆. Do you think you'll have time to address any of the comments from @alice above? If not, it might be worth rebasing against master so they can be looked at in another round of work in the future.

ricksbrown:
@addyosmani Thanks for the encouragement 😄 I dropped this for a while. I shall take a look at it pronto.

addyosmani:
@alice were there any other pieces of feedback you wanted addressed before we landed this?

addyosmani:
Currently have some CI failures for this PR I'm investigating. Going to restart it for this PR and see if we can go green.

addyosmani:
Landing these changes now. We can address further feedback in a follow-up PR as time allows.

src/js/AuditRule.js
dran7505: This diff

PR #267, opened 2015-12-16
fix: escape special characters to be used in a query selector string

joscha:
Currently when the CSS selector is generated for an ID that contains special characters (let's say dots) these are not escaped, resulting in an incorrect query selector.
Example `<div id="id.with.dots">` should yield `#id\.with\.dots`. Instead it yields `#id.with.dots` which searches for an element with the ID `id` and two classes `with` and `dots`.

This PR fixes this and adds a test.

### DIFF
--- a/src/js/AccessibilityUtils.js
+++ b/src/js/AccessibilityUtils.js
@@ -922,6 +922,17 @@ axs.utils.namedValues = function(obj) {
     return values;
 };
 
+/**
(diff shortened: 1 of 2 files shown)


joscha:
anything else I can do to help this get merged?

alice:
Sorry for being so slow on this! Looks good, thanks for doing this.

PR #311, opened 2016-10-27
Update `composedParentNode` for Shadow DOM v1.

### LINKED ISSUE #312: `axs.dom.composedParentNode` broken in Shadow DOM v1
`axs.dom.composedParentNode` should use `.assignedSlot` instead of `.getDestinationInsertionPoints()` when in Shadow DOM v1.

I noticed an element I was working on kept failing [`AX_FOCUS_01`](https://github.com/GoogleChrome/accessibility-developer-tools/blob/b5fd54a6629125b582d10f7b183c7a23151d0894/src/audits/FocusableElementNotVisibleAndNotAriaHidden.js#L53) and found that `elementIsVisible` was returning false because it wasn't able to figure out that an element it found above the one passed to it was actually a descendant and not just arbitrarily obscuring it.

bicknellr:
This PR updates `axs.dom.composedParentNode` to use `assignedSlot` (new in Shadow DOM v1, [link](https://github.com/w3c/webcomponents/commit/224f6b8e1e48f5db0a592ec8fb5162770062be24)) if available and doesn't attempt to call `getDestinationInsertionPoints` (removed) if it isn't available.

Some of the tests specifically check the behavior of content elements with a set `select` attribute; I changed these to use slots with `name` attributes and added `slot` attributes to the relevant nodes outside the shadow tree to get the same 'distributed vs. not distributed' behavior. It might be worth renaming the tests since they're not about content elements anymore.

(fixes #312)

### DIFF
--- a/src/js/DOMUtils.js
+++ b/src/js/DOMUtils.js
@@ -70,9 +70,15 @@ axs.dom.composedParentNode = function(node) {
     if (!parentNode.shadowRoot)
         return parentNode;
 
-    var insertionPoints = node.getDestinationInsertionPoints();
-    if (insertionPoints.length > 0)
-        return axs.dom.composedParentNode(insertionPoints[insertionPoints.length - 1]);
+    var assignedSlot = node.assignedSlot;
+    if (assignedSlot)
+        return axs.dom.composedParentNode(assignedSlot);
+
+    if ('getDestinationInsertionPoints' in node) {
+        var insertionPoints = node.getDestinationInsertionPoints();
+        if (insertionPoints.length > 0)
+            return axs.dom.composedParentNode(insertionPoints[insertionPoints.length - 1]);
+    }
 
     return null;
 };
--- a/src/js/DOMUtils.js
+++ b/src/js/DOMUtils.js
@@ -70,13 +70,15 @@ axs.dom.composedParentNode = function(node) {
     if (!parentNode.shadowRoot)
         return parentNode;
 
+    // Shadow DOM v1
     if (node.nodeType === Node.ELEMENT_NODE || node.nodeType === Node.TEXT_NODE) {
       var assignedSlot = node.assignedSlot;
-      if (assignedSlot)
+      if (HTMLSlotElement && assignedSlot instanceof HTMLSlotElement)
           return axs.dom.composedParentNode(assignedSlot);
     }
 
-    if ('getDestinationInsertionPoints' in node) {
+    // Shadow DOM v0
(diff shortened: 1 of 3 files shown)


bicknellr:
After cleaning, it looks like this causes a closure compiler warning about `assignedSlot` never being defined on `Node`. Maybe I should try to get Shadow DOM v1 stuff into closure first? :/

alice:
Sorry for not looking at this sooner!

You can patch around this locally by adding the following in `externs.js`:
```js
/**
 * @constructor
 * @extends {HTMLElement}
 */
function HTMLSlotElement() {}

/**
 * @return {?HTMLSlotElement}
 */
HTMLElement.prototype.assignedSlot = function() {};
```

And you might want to make a PR on https://github.com/google/closure-compiler/blob/master/externs/browser/html5.js to add the same lines where appropriate.

alice:
(Meanwhile, https://github.com/GoogleChrome/accessibility-developer-tools/pull/327 updates us to use the google-released closure compiler which should stay more up to date)

bicknellr:
Updated with your patch - no more closure errors!

bicknellr:
Oh, I still need to check for the right node types...

src/js/externs/externs.js
```diff
@@ -30,3 +30,14 @@ ShadowRoot.prototype.olderShadowRoot;
...
+/**
+ * @constructor
+ * @extends {HTMLElement}
+ */
+function HTMLSlotElement() {}
+
+/**
+ * @return {?HTMLSlotElement}
+ */
+HTMLElement.prototype.assignedSlot = function() {};
```
bicknellr: `HTMLElement` -> `Element`

src/js/DOMUtils.js
```diff
@@ -70,9 +70,17 @@ axs.dom.composedParentNode = function(node) {
...
-    var insertionPoints = node.getDestinationInsertionPoints();
-    if (insertionPoints.length > 0)
-        return axs.dom.composedParentNode(insertionPoints[insertionPoints.length - 1]);
+    if (node.nodeType === Node.ELEMENT_NODE || node.nodeType === Node.TEXT_NODE) {
+      var assignedSlot = node.assignedSlot;
+      if (assignedSlot)
+          return axs.dom.composedParentNode(assignedSlot);
+    }
+
+    if ('getDestinationInsertionPoints' in node) {
```
alice: How come we're doing an 'in' check here, but not for `assignedSlot`?

bicknellr:
No problem!

PR #338, opened 2017-02-16
handle nodes hosted in a <template>

### LINKED ISSUE #337: hasDirectTextDescendantTreeWalker fails when encountering a DocumentFragment in IE11
https://github.com/GoogleChrome/accessibility-developer-tools/blob/master/src/js/Properties.js#L178

`hasDirectTextDescendantTreeWalker` assumes the `parentNode` will have `tagName` defined, but that's not the case for DocumentFragments.

valdrinkoshi:
Fixes #337.
@alice FYI

### DIFF
--- a/src/js/Properties.js
+++ b/src/js/Properties.js
@@ -175,8 +175,9 @@ axs.properties.hasDirectTextDescendant = function(element) {
         while (treeWalker.nextNode()) {
             var resultElement = treeWalker.currentNode;
             var parent = resultElement.parentNode;
-            // parent might not be an Element (e.g. DocumentFragment).
-            var tagName = (parent.tagName || '').toLowerCase();
+            // Handle elements hosted in <template>.content.
+            parent = parent.host || parent;
+            var tagName = parent.tagName.toLowerCase();
             var value = resultElement.nodeValue.trim();
             if (value && tagName !== 'script' && element !== resultElement)
                 return true;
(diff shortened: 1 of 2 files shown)


src/js/Properties.js
```diff
@@ -175,7 +175,8 @@ axs.properties.hasDirectTextDescendant = function(element) {
         while (treeWalker.nextNode()) {
             var resultElement = treeWalker.currentNode;
             var parent = resultElement.parentNode;
-            var tagName = parent.tagName.toLowerCase();
+            // parent might not be an Element (e.g. DocumentFragment).
+            var tagName = (parent.tagName || '').toLowerCase();
```
alice: Would it make sense to just switch to parentElement on the previous line?

valdrinkoshi:
@alice I've realized the issue happens only if a node is hosted in a <template>, because if we use a DocumentFragment to host nodes and then add the fragment to the main document, the parent of each node will be the new parent, not the DocumentFragment.
This is not the case for <template>:
```js
var frag = document.createDocumentFragment();
frag.appendChild(document.createElement('div'));
document.body.appendChild(frag); // the <div>.parentNode is <body>

var templ = document.createElement('template');
templ.content.appendChild(document.createElement('div'));
document.body.appendChild(templ); // the <div>.parentNode is <template>.content!
```

I've updated the test and the implementation to handle specifically <template>, PTAL 🍰

addyosmani:
LGTM2

PR #341, opened 2017-03-08
specify support of ShadowRoot

valdrinkoshi:
https://github.com/GoogleChrome/accessibility-developer-tools/pull/338 actually provided support for elements hosted in a shadow root, not a `<template>`. This PR makes it more clear & updates the tests accordingly.
 
When an element is hosted in `<template>.content`, its `.parentNode` will be a `DocumentFragment`, and the fragment's `.host` will be `undefined`.
When an element is hosted in a `shadowRoot`, its `.parentNode` will be a `DocumentFragment`, and the fragment's `.host` will be the shadowRoot owner element 🎉


valdrinkoshi:
@alice @addyosmani FYI. This is just to have more robust tests 💪

PR #343, opened 2017-04-25
updating for aria 1.1 - WIP

stevefaulkner:
updated application, rowgroup, rowheader, columnheader 
added switch, none, cell, table, feed


stevefaulkner:
@alice  can you take a squizz and tell me if i am on right track?

Issue #345, opened 2017-04-28
axs.utils.isLargeFont returns incorrect values with non-integer fonts


timmhayes:
The `axs.utils.isLargeFont()` function fails on font sizes with float values. 

I saw this first when testing a value of 10pt, which gets a computed value of 13.3333px in Chrome. The current function evaluates this incorrectly as a **_large_** font with a value of "3333px".  Or conversely, a font value of 30.1px evaluates incorrectly as a **_small_** font with a value of "1px"

See a working example here: http://jsbin.com/cehiben/edit?html,output

A small chunk of testable code extracted from isLargeFont() shows the issue: 

```
 var fontSize = "13.3333px"
 var matches = fontSize.match(/(\d+)px/);
 if (matches) {
   var fontSizePx = parseInt(matches[1], 10); // parses as 3333, not 13
   console.log(fontSizePx)
 }
```
I'll submit a pull request with a suggested fix.

This is also causing a failure in Khan Academy's [tota11y ](http://khan.github.io/tota11y/) toolkit.

PR #346, opened 2017-04-28
Fix axs.utils.isLargeFont float issue (#345)

timmhayes:
This fix should allow the `isLargeFont ` function to correctly evaluate font sizes with decimals. 

As it is now, a font sizes of 10pt can be evaluated as a **_large_** font because Chrome returns it as having a size of 13.333px, which the RegEx returns as 3333.


timmhayes:
I signed the CLA

PR #358, opened 2017-11-16
Resolve circular dependency and avoid using reserved keyword 'abstract'

alice:
@robdodson PTAL


robdodson:
LGTM :P

Joyboy83:
alice:master merging  accesss

